PR #19, opened 2023-09-17
fix: reuse msgspec Encoder and Decoder instances

guacs:

### Pull Request Checklist

- [ ] Pre-Commit Checks were ran and passed

### Description

- This PR reuses the `Encoder` and `Decoder` instances when using `msgspec` as the JSON serializer/deserializer.

### DIFF
--- a/advanced_alchemy/config/engine.py
+++ b/advanced_alchemy/config/engine.py
@@ -15,23 +15,25 @@ if TYPE_CHECKING:
     from advanced_alchemy.config.types import EmptyType
 
 try:
-    from msgspec.json import decode as decode_json
-    from msgspec.json import encode as _encode_json
+    from msgspec.json import Decoder, Encoder
+
+    encoder, decoder = Encoder(), Decoder()
+    decode_json = decoder.decode
 
     def encode_json(data: Any) -> str:
-        return _encode_json(data).decode("utf-8")
+        return encoder.encode(data).decode("utf-8")
 
 except ImportError:
     try:
-        from orjson import dumps as _encode_json  # type: ignore[no-redef]
+        from orjson import dumps as _encode_json
         from orjson import loads as decode_json  # type: ignore[no-redef]
 
         def encode_json(data: Any) -> str:
-            return _encode_json(data).decode("utf-8")
+            return _encode_json(data).decode("utf-8")  # type: ignore[no-any-return]
 
     except ImportError:
         from json import dumps as encode_json  # type: ignore[assignment]
-        from json import loads as decode_json  # type: ignore[no-redef]
+        from json import loads as decode_json  # type: ignore[assignment]
 
 __all__ = ("EngineConfig",)
 


guacs:
As per the [msgspec docs](https://jcristharif.com/msgspec/perf-tips.html#reuse-encoders-decoders), it's a good idea to reuse the Encoder/Decoder for performance.

advanced_alchemy/config/engine.py
```diff
@@ -15,23 +15,25 @@
...
 
 except ImportError:
     try:
-        from orjson import dumps as _encode_json  # type: ignore[no-redef]
+        from orjson import dumps as _encode_json
         from orjson import loads as decode_json  # type: ignore[no-redef]
 
         def encode_json(data: Any) -> str:
-            return _encode_json(data).decode("utf-8")
+            return _encode_json(data).decode("utf-8")  # type: ignore[no-any-return]
```
guacs, line 32: I have no idea why `mypy` kept complaining this is returning `Any`. `pyright` was able to resolve it properly, but `mypy` was not.

cofin (approved):

cofin:
Thanks for the help finishing this up

provinzkraut:
iirc @jcrist mentioned somewhere that the difference of doing this should be negligible if you're not specifying a type for the `Decoder`.

jcrist:
This is correct. There used to be additional allocation-avoiding benefits, but these were removed in https://github.com/jcrist/msgspec/pull/402 to make the encoders/decoders threadsafe. Creating a `Decoder` beforehand is now only more efficient if you're passing in a type (to avoid re-parsing the type every call, which *can* be expensive for non-struct-types) or if passing in other configuration like `dec_hook` (although this is likely negligible for most users).

guacs:
> This is correct. There used to be additional allocation-avoiding benefits, but these were removed in [jcrist/msgspec#402](https://github.com/jcrist/msgspec/pull/402) to make the encoders/decoders threadsafe. Creating a `Decoder` beforehand is now only more efficient if you're passing in a type (to avoid re-parsing the type every call, which _can_ be expensive for non-struct-types) or if passing in other configuration like `dec_hook` (although this is likely negligible for most users).

Is this the same case for `Encoder` as well?

jcrist:
Yes, except that encoding is generally a faster process than decoding, so the constant time taken to process kwargs will be a higher fraction of the time per-call. So there might be more of a benefit for doing kwarg processing once when creating an `Encoder` rather than every call to `encode`.

In the case where you're passing no kwarg options, then there should be negligible benefit to creating an `Encoder` over just calling the top-level `encode` function.

guacs:
> so the constant time taken to process kwargs will be a higher fraction of the time per-call

Oh, this is something I didn't consider at all. Basically, we could just prevent the call to `Encoder_init` right?

What about the memory allocation/deallocation though? Are they not expensive or are they negligible as well?

PR #39, opened 2023-09-28
msgspec missing from dependencies

sfermigier:
msgspec is imported from the library, but currently declared as optional in pyproject.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
 dependencies = [
   "sqlalchemy>=2.0.20",
   "alembic>=1.12.0",
+  "msgspec>=0.18.2",
   "typing_extensions",
 ]
 description = "Ready-to-go SQLAlchemy concoctions."


guacs:
I don't think adding `msgspec` as the dependency is the solution. Instead, we should move this [part](https://github.com/jolt-org/advanced-alchemy/blob/3126ac61b178666b7bebaa4a8d3e3312bc28dcfd/advanced_alchemy/config/engine.py#L17-L36) into a file of its own and import the `decode_json`, `encode_json` from there.

cofin:
> I don't think adding `msgspec` as the dependency is the solution. Instead, we should move this [part](https://github.com/jolt-org/advanced-alchemy/blob/3126ac61b178666b7bebaa4a8d3e3312bc28dcfd/advanced_alchemy/config/engine.py#L17-L36) into a file of its own and import the `decode_json`, `encode_json` from there.

I think this is the right approach here as well. The reason being is that `fastapi`, `starlette`, and `sanic` don't have a `msgspec` dependency, so it probably should be optional for those use cases.

sfermigier:
OK.

PR #45, opened 2023-10-03
feat: service layer abstraction

cofin:
feat: adds a service layer for the repositories that can be extended


advanced_alchemy/service/_async.py
```diff
@@ -0,0 +1,526 @@
...
+
+        Returns:
+            Representation of created instance.
+        """
+        data = await self.to_model(data, "create")
+        return await self.repository.add(data)
+
+    async def create_many(
+        self,
+        data: list[ModelT | dict[str, Any]] | list[dict[str, Any]] | list[ModelT],
```
Alc-Alc, line 94: Can this and the other occurrences be replaced with `ModelDictListT` alias defined in `advanced_alchemy/service/typing.py`?

```suggestion
        data: ModelDictListT
```

Alc-Alc (approved):

advanced_alchemy/service/_async.py
cofin: It ended up complaining in a few other places on the repository then.  I'm sure there's a way to clean these up though.

.github/workflows/ci.yaml
```diff
@@ -45,6 +45,19 @@ jobs:
 
       - name: Install Microsoft ODBC Drivers
         run: sudo ACCEPT_EULA=Y apt-get install msodbcsql18 -y
+      - name: Free additional space
```
provinzkraut, line 48: Just curious, why is this needed?

provinzkraut (approved):

.github/workflows/ci.yaml
cofin: We've had quite a few failures because we go beyond the 20gb or so that Github allocates

provinzkraut: Oh :eyes:

PR #46, opened 2023-10-04
feat: custom encrypted field type

cofin:
SQLAlchemy Encrypted field with support for multiple encryption backends.


cofin:
This has more changes than I expected.  Closing and will re-open with only the encrypted field

PR #47, opened 2023-10-04
feat: SQLAlchemy encrypted field

cofin:
This PR adds an encrypted field for SQLAlchemy


advanced_alchemy/types.py
```diff
@@ -174,3 +175,42 @@ def process_result_value(self, value: datetime.datetime | None, dialect: Dialect
...
+    def column_expression(self, column: Any) -> Any:
+        return self._handle_encrypted_data(column)
+
+    def _handle_encrypted_data(self, column: Any) -> str:
+        if self.backend == "pgcrypto":
+            return sql_func.pgp_sym_encrypt(column, self.passphrase)  # type: ignore[return-value]
+        if self.backend == "tink":
+            raise NotImplementedError
+        if self.backend == "cryptography":
+            raise NotImplementedError
```
provinzkraut, line 227: Wouldn't it make more sense to not have these in the first place if they're currently not implemented?

Sounds less confusing to me.

advanced_alchemy/types.py
```diff
@@ -174,3 +175,42 @@ def process_result_value(self, value: datetime.datetime | None, dialect: Dialect
...
+        return None  # type: ignore[unreachable]
```
provinzkraut, line 228: Could add a
```python
typing.assert_never()
```
to have mypy assert the check is exhaustive.

provinzkraut (approved):

advanced_alchemy/types.py
cofin: I expect to have them before this PR is merged.  I was really trying to set up the structure so I could think through the whole thing.

provinzkraut: Oh. Sorry for the slightly overzealous comments then :slightly_smiling_face:

provinzkraut:
@cofin is this not good to go?

cofin:
> @cofin is this not good to go?

Yes.  The PGCrypto variant is ready to go.  I was going to finish up the cryptography and tink backends before merging, but maybe that should just get split into a separate PR.

provinzkraut:
> but maybe that should just get split into a separate PR.

Sounds good

PR #50, opened 2023-10-09
feat: configurable autocommit handler factory function for Litestar plugin

geeshta:

### Pull Request Checklist

### Description
Introduction of a new factory function `autocommit_handler` to create handlers that supports:
- handling all redirects (3XX) with performing a commit
- specifying custom sets of status codes for commit and/or rollback

This might be desirable for MPAs, where it is quite common to use the [PRG pattern](https://en.wikipedia.org/wiki/Post/Redirect/Get) - after a successful form submission, a redirect response is issued to prevent multiple submission (by for example reloading the page).
It also gives more flexibility with defining when the transaction should be rolled back or committed.

~~This change is breaking for applications, that rely on the old behaviour, i.e. having the transaction rolled back upon any redirect.~~

### DIFF
--- a/advanced_alchemy/extensions/litestar/plugins/init/config/asyncio.py
+++ b/advanced_alchemy/extensions/litestar/plugins/init/config/asyncio.py
@@ -55,9 +55,9 @@ def autocommit_handler_maker(
 ) -> Callable[[Message, Scope], Coroutine[Any, Any, None]]:
     """Set up the handler to issue a transactin commit or rollback based on specified status codes
     Args:
-        commit_on_redirect: issue a commit when the response status is a redirect (3XX)
-        extra_commit_statuses: a set of additional status codes that trigger a commit
-        extra_rollback_statuses: a set of additional status codes that trigger a rollback
+        commit_on_redirect: Issue a commit when the response status is a redirect (``3XX``)
+        extra_commit_statuses: A set of additional status codes that trigger a commit
+        extra_rollback_statuses: A set of additional status codes that trigger a rollback
 
     Returns:
         The handler callable
--- a/advanced_alchemy/extensions/litestar/plugins/init/config/asyncio.py
+++ b/advanced_alchemy/extensions/litestar/plugins/init/config/asyncio.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Callable, cast
 
 from litestar.constants import HTTP_RESPONSE_START
-from litestar.status_codes import HTTP_200_OK, HTTP_300_MULTIPLE_CHOICES
+from litestar.status_codes import HTTP_200_OK, HTTP_300_MULTIPLE_CHOICES, HTTP_302_FOUND, HTTP_303_SEE_OTHER
 from litestar.utils import delete_litestar_scope_state, get_litestar_scope_state, set_litestar_scope_state
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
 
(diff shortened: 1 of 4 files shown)


cofin (approved):

provinzkraut:
> This change is breaking for applications, that rely on the old behaviour, i.e. having the transaction rolled back upon any redirect.

Maybe this should be configurable instead then?

geeshta:
```python
class AutoCommitHandler:
    __slots__ = ("commit_on_redirect", "extra_commit_statuses", "extra_rollback_statuses")

    def __init__(
        self,
        commit_on_redirect: bool = False,
        extra_commit_statuses: set[int] = set(),
        extra_rollback_statuses: set[int] = set(),
    ) -> None:
        """Set up the handler to issue a transactin commit or rollback based on specified status codes
        Args:
            commit_in_redirect: issue a commit when the response status is a redirect (3XX)
            extra_commit_statuses: a set of additional status codes that trigger a commit
            extra_rollback_statuses: a set of additional status codes that trigger a rollback

        Returns:
            None
        """
        if len(extra_commit_statuses & extra_rollback_statuses) > 0:
            raise ValueError("Rollback extra statuses and commit extra statuses must not contain common status codes")

        self.commit_on_redirect = commit_on_redirect
        self.extra_commit_statuses = extra_commit_statuses
        self.extra_rollback_statuses = extra_rollback_statuses

    async def __call__(self, message: Message, scope: Scope) -> None:
        """Handle commit/rollback, closing and cleaning up sessions before sending.

        Args:
            message: ASGI-``Message``
            scope: An ASGI-``Scope``

        Returns:
            None
        """
        session = cast("AsyncSession | None", get_litestar_scope_state(scope, SESSION_SCOPE_KEY))
        try:
            if session is not None and message["type"] == HTTP_RESPONSE_START:
                commit_range = range(200, 300 if not self.commit_on_redirect else 400)
                if (
                    cast("HTTPResponseStartEvent", message)["status"] in commit_range
                    or message["status"] in self.extra_commit_statuses
                ) and message["status"] not in self.extra_rollback_statuses:
                    await session.commit()
                else:
                    await session.rollback()
        finally:
            if session and message["type"] in SESSION_TERMINUS_ASGI_EVENTS:
                await session.close()
                delete_litestar_scope_state(scope, SESSION_SCOPE_KEY)


autocommit_before_send_handler = AutoCommitHandler()
```

geeshta:
What do you think of this? This should introduce breaking changes if used in the standard way. It would probably cause some typing changes which need to be fixed though. The usage would be like this:
```python
autocommit_redirect_handler = AutoCommitHandler(commit_on_redirect=True)

custom_autocommit_handler = AutoCommitHandler(extra_commit_statuses={HTTP_302_FOUND, HTTP_303_SEE_OTHER})
```

advanced_alchemy/extensions/litestar/plugins/init/config/asyncio.py
```diff
@@ -49,27 +48,60 @@ async def default_before_send_handler(message: Message, scope: Scope) -> None:
...
+            message: ASGI-``Message``
+            scope: An ASGI-``Scope``
+
+        Returns:
+            None
+        """
+        session = cast("AsyncSession | None", get_litestar_scope_state(scope, SESSION_SCOPE_KEY))
+        try:
+            if session is not None and message["type"] == HTTP_RESPONSE_START:
+                commit_range = range(200, 300 if not commit_on_redirect else 400)
```
peterschutt, line 88: This can be moved outside of the closure.

geeshta:
- Renamed to `autocommit_handler_maker` to distinguish it from the actual handlers
- Made as a factory function to prevent any typing issues with callable classes
- Wrote tests
- The original handler function is identical and backwards compatible

advanced_alchemy/extensions/litestar/plugins/init/config/asyncio.py
```diff
@@ -49,27 +48,61 @@ async def default_before_send_handler(message: Message, scope: Scope) -> None:
...
+    extra_commit_statuses: set[int] | None = None,
+    extra_rollback_statuses: set[int] | None = None,
+) -> Callable[[Message, Scope], Coroutine[Any, Any, None]]:
+    """Set up the handler to issue a transactin commit or rollback based on specified status codes
     Args:
-        message: ASGI-``Message``
-        scope: An ASGI-``Scope``
+        commit_on_redirect: issue a commit when the response status is a redirect (3XX)
+        extra_commit_statuses: a set of additional status codes that trigger a commit
+        extra_rollback_statuses: a set of additional status codes that trigger a rollback
```
JacobCoffee, line 60:
```suggestion
        commit_on_redirect: Issue a commit when the response status is a redirect (``3XX``)
        extra_commit_statuses: A set of additional status codes that trigger a commit
        extra_rollback_statuses: A set of additional status codes that trigger a rollback
```

JacobCoffee (approved):

cofin (approved):
Thanks @geeshta!

LGTM

PR #52, opened 2023-10-14
fix: CollectionFilter returns all entries if values is empty

### LINKED ISSUE #51: Bug: CollectionFilter returns all entries if values is empty
### Description

Basically what the title says, when initializing a CollectionFilter with an empty list, the resulting query returns all entries instead of no entries, which I'd expect. 
I know it's a special case and needs to be handled differently, but this seems very inconsistent behaviour. 
Here's the specific line which leads to that bug (currently in advanced-alchemy project, although the code is unchanged since it was part of litestar)
https://github.com/jolt-org/advanced-alchemy/blob/46d3e7acbc7a391b4bab06fe7e64f3d45826270a/advanced_alchemy/repository/_async.py#L1109

Issue was first raised on the litestar discord
The bug is replicated in the sync repo as well

_No response_

### MCVE

_No response_

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Jolt Project Version

master

### Platform

- [X] Linux
- [X] Mac
- [X] Windows

---

## Funding
* If you would like to see an issue prioritized, make a pledge towards it!
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/jolt-org/advanced-alchemy/issues/51">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/jolt-org/advanced-alchemy/issues/51/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

cofin:
Fixes #51

Bug: CollectionFilter returns all entries if values is empty

a simple `1=-1` is appended into the `where` clause when an empty list is passed into the `in` statement.

### DIFF
--- a/tests/integration/test_repository.py
+++ b/tests/integration/test_repository.py
@@ -1033,6 +1033,9 @@ async def test_repo_filter_collection(
     existing_obj = await maybe_async(author_repo.list(CollectionFilter(field_name="id", values=[second_author_id])))
     assert existing_obj[0].name == "Leo Tolstoy"
 
+    no_obj = await maybe_async(author_repo.list(CollectionFilter(field_name="id", values=[])))
+    assert no_obj is None
+
 
 async def test_repo_filter_not_in_collection(
     author_repo: AuthorRepository,
--- a/tests/integration/test_repository.py
+++ b/tests/integration/test_repository.py
@@ -1033,6 +1033,10 @@ async def test_repo_filter_collection(
     existing_obj = await maybe_async(author_repo.list(CollectionFilter(field_name="id", values=[second_author_id])))
     assert existing_obj[0].name == "Leo Tolstoy"
 
+
+async def test_repo_filter_no_obj_collection(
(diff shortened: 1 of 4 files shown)


peterschutt (approved):

tests/integration/test_repository.py
```diff
@@ -1033,6 +1033,9 @@ async def test_repo_filter_collection(
     existing_obj = await maybe_async(author_repo.list(CollectionFilter(field_name="id", values=[second_author_id])))
     assert existing_obj[0].name == "Leo Tolstoy"
 
+    no_obj = await maybe_async(author_repo.list(CollectionFilter(field_name="id", values=[])))
```
provinzkraut, line 1036: I think it would be better to have this as its own test function

cofin: Cool.  I'll update this morning before getting out a patch release.

JacobCoffee (approved):

PR #59, opened 2023-10-17
feat: use `ANY` over `IN` for certain engines

cofin:


.github/workflows/ci.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
           python-version: "3.11"
 
       - name: Install Pre-Commit
-        run: python -m pip install pre-commit && pre-commit install
+        run: python -m pip install pre-commit
```
peterschutt, line 20: Are these changes related to the PR?

advanced_alchemy/filters.py
```diff
@@ -58,8 +58,8 @@ class CollectionFilter(Generic[T]):
 
     field_name: str
     """Name of the model attribute to filter on."""
-    values: abc.Collection[T]
-    """Values for ``IN`` clause."""
+    values: abc.Collection[T] | None
+    """Values for ``IN`` clause. An empty list will return an empty result set, however, if ``None``, the filter is not applied to the query, and all rows are returned. """
```
peterschutt, line 62: Nit, but should we wrap this line?

advanced_alchemy/repository/_sync.py
```diff
@@ -54,6 +54,9 @@ class SQLAlchemySyncRepository(Generic[ModelT]):
     model_type: type[ModelT]
     id_attribute: Any = "id"
     match_fields: list[str] | str | None = None
+    _prefer_any_clause: bool = False
+    prefer_any_clause_engines: tuple[str] | None = ("postgresql",)
```
peterschutt, line 58: Is the union to `None` necessary here? Overriding sub-classes could set it to `()` if they wanted to override for some reason.

advanced_alchemy/repository/_sync.py
```diff
@@ -94,6 +97,9 @@ def __init__(
             msg = "Session improperly configure"
             raise ValueError(msg)
         self._dialect = self.session.bind.dialect
+        self._prefer_any_clause = any(
+            self._dialect.name == engine_type for engine_type in self.prefer_any_clause_engines or ()
+        )
```
peterschutt, line 102: What if `prefer_any_clause_engines` was a set?

```suggestion
        self._prefer_any_clause = self._dialect.name in self.prefer_any_clause_engines
```

peterschutt (commented):
Nice stuff overall, but from a meta perspective, it feels like this is really 3 PRs rolled into one.

1. some build stuff
2. the ANY over IN stuff
3. the changes for `None` arguments to collection filters.

I think it would make it much better for our future selves, and newcomers if we keep to more focused PRs with adequate descriptions of the what and why.

advanced_alchemy/repository/_sync.py
cofin: I had it as a set initially, but I wa getting errors about it being unhashable.  That was the reason for it to go to a `tuple`.  A `set` actually made more sense to me.

advanced_alchemy/repository/_sync.py
cofin: It's not necessary.  I thought it might be more intuitive to also be able to set it to `None` to disable it.  Ideally, the user shouldn't be overriding this anyway.

.github/workflows/ci.yaml
cofin: Nope... I'll split it up.

cofin:
> Nice stuff overall, but from a meta perspective, it feels like this is really 3 PRs rolled into one.
> 
> 1. some build stuff
> 2. the ANY over IN stuff
> 3. the changes for `None` arguments to collection filters.
> 
> I think it would make it much better for our future selves, and newcomers if we keep to more focused PRs with adequate descriptions of the what and why.

I thought I might be able to squeeze a few things in :) 

However, I do agree.  I'll split out the other changes.

cofin:
Rather than fight these conflicts, i'm going to close and re-open this.

advanced_alchemy/repository/_sync.py
peterschutt: I see we retrieve the dialect from the session:

https://github.com/jolt-org/advanced-alchemy/blob/48265257ccd810960f51b4ad5bd83e14d4abd469/advanced_alchemy/repository/_async.py#L95C45-L95C45

`Session.bind` is typed `Optional[Union[Engine, Connection]]` [ref](https://github.com/sqlalchemy/sqlalchemy/blob/efe4025c068d6746637db362245ab88ce7dfe785/lib/sqlalchemy/orm/session.py#L1429C46-L1429C46).

`Engine.dialect` is just `Dialect` [ref](https://github.com/sqlalchemy/sqlalchemy/blob/efe4025c068d6746637db362245ab88ce7dfe785/lib/sqlalchemy/engine/base.py#L2935), and `Connection.dialect` should be same [ref](https://github.com/sqlalchemy/sqlalchemy/blob/efe4025c068d6746637db362245ab88ce7dfe785/lib/sqlalchemy/engine/base.py#L141C18-L141C18).

`Dialect.name` is typed `str` [ref](https://github.com/sqlalchemy/sqlalchemy/blob/efe4025c068d6746637db362245ab88ce7dfe785/lib/sqlalchemy/engine/interfaces.py#L664). So hard to see where the unhashable error might be coming from. You could try `reveal_type(self._diaclect.name)` and `reveal_type(self.prefer_any_clause_engines)` in the scope that the error was originating from to see if that turns up anything.

peterschutt:
> > Nice stuff overall, but from a meta perspective, it feels like this is really 3 PRs rolled into one.
> > 
> > 1. some build stuff
> > 2. the ANY over IN stuff
> > 3. the changes for `None` arguments to collection filters.
> > 
> > I think it would make it much better for our future selves, and newcomers if we keep to more focused PRs with adequate descriptions of the what and why.
> 
> I thought I might be able to squeeze a few things in :)
> 
> However, I do agree. I'll split out the other changes.

Sorry to be a pain!

PR #61, opened 2023-10-18
chore: optimized test build & coverage collection

cofin:


.github/workflows/ci.yaml
```diff
@@ -64,18 +60,39 @@ jobs:
...
+      - name: Install base libraries
+        # install a few core libraries for testing and building
+        run: pip install nodeenv cython setuptools mypy cython pip --upgrade --quiet --user
       - name: Install dependencies
-        run: pdm install
-
-      - name: Test
-        run: pdm run pytest -m "" -n auto
+        run: pdm install -G:all
+      - name: Test with Coverage
```
guacs, line 85: Does this need to be run for all python versions?

guacs (approved):

.github/workflows/ci.yaml
guacs: How about something like [this](https://github.com/litestar-org/polyfactory/blob/e1f7a47c6e3c979f8f43ce6ee612f417d2f3c03b/.github/workflows/ci.yaml#L84-L86)? Or the litestar one which also only tests with coverage for 3.12 and pydantic v2 ([reference](https://github.com/litestar-org/litestar/blob/3a2089a96483652f8cadb4292eb9ae42a28f4db6/.github/workflows/ci.yml#L43)). It does have a more elaborate setup than the polyfactory one though.

cofin: Thanks.  I've updated it to work like the other repos.

peterschutt (approved):

PR #62, opened 2023-10-18
fix: better handle empty collection filters

cofin:
Currently, [this](https://github.com/cofin/litestar-fullstack/blob/main/src/app/lib/dependencies.py#L169) is how you can inject these filters in your app.

When using the `id_filter` dependency on it's own, you have to have an additional not-null check before passing it into the repository.  

This change handles that and allows you to pass in all fitlers into the repository function without checking their nullability.


tests/unit/test_repository.py
```diff
@@ -644,6 +644,21 @@ async def test_sqlalchemy_repo_list_with_collection_filter(
...
+    mock_repo_execute: AnyMock,
+    mocker: MockerFixture,
+) -> None:
+    """Test behavior of list operation given CollectionFilter."""
+    field_name = "id"
+    mock_repo_execute.return_value = MagicMock()
+    mock_repo.statement.where.return_value = mock_repo.statement
+    mocker.patch.object(mock_repo, "_filter_in_collection", return_value=mock_repo.statement)
+    await maybe_async(mock_repo.list(CollectionFilter(field_name, None)))
+    assert mock_repo._filter_in_collection.call_count == 0
```
peterschutt, line 659: This can also be `mock_repo._filter_in_collection.assert_not_called()`

tests/unit/test_repository.py
```diff
@@ -678,6 +693,21 @@ async def test_sqlalchemy_repo_list_with_not_in_collection_filter(
...
+    mock_repo_execute: AnyMock,
+    mocker: MockerFixture,
+) -> None:
+    """Test behavior of list operation given CollectionFilter."""
+    field_name = "id"
+    mock_repo_execute.return_value = MagicMock()
+    mock_repo.statement.where.return_value = mock_repo.statement
+    mocker.patch.object(mock_repo, "_filter_not_in_collection", return_value=mock_repo.statement)
+    await maybe_async(mock_repo.list(NotInCollectionFilter(field_name, None)))
+    assert mock_repo._filter_not_in_collection.call_count == 0
```
peterschutt, line 708: Same

peterschutt (approved):

PR #71, opened 2023-10-23
feat: verify the service has an identity before updating.

cofin:
This PR raises a repository error when you call the service `update` method without an `item_id` and without a value for the ID attribute.

This ensures you are operating with data that can actually be updated.

### DIFF
--- a/advanced_alchemy/service/_async.py
+++ b/advanced_alchemy/service/_async.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any, Generic, Iterable, cast
 
 from sqlalchemy.orm import InstrumentedAttribute
 
+from advanced_alchemy.exceptions import RepositoryError
 from advanced_alchemy.repository._util import model_from_dict
 from advanced_alchemy.repository.typing import ModelT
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_repository.py
```diff
@@ -1550,3 +1550,8 @@ async def test_repo_get_or_create_deprecation(author_repo: AuthorRepository, fir
         existing_obj, existing_created = await maybe_async(author_repo.get_or_create(name="Agatha Christie"))
         assert existing_obj.id == first_author_id
         assert existing_created is False
+
+
+async def test_service_update_no_pk(author_service: AuthorService, first_author_id: Any) -> None:
```
peterschutt, line 1555:
```suggestion
async def test_service_update_no_pk(author_service: AuthorService) -> None:
```

advanced_alchemy/service/_async.py
```diff
@@ -316,6 +317,11 @@ async def update(
         data = await self.to_model(data, "update")
         if isinstance(id_attribute, InstrumentedAttribute):
             id_attribute = cast("str", id_attribute.description)
+        if item_id is None and self.repository.get_id_attribute_value(item=data, id_attribute=id_attribute) is None:
+            msg = f"Could not identify ID attribute value.  One of the following is required: ``item_id`` or ``data.{id_attribute or self.repository.id_attribute}``"
```
peterschutt, line 321:
```suggestion
            msg = (
                "Could not identify ID attribute value.  One of the following is required: "
                f"``item_id`` or ``data.{id_attribute or self.repository.id_attribute}``"
            )
```

advanced_alchemy/service/_sync.py
```diff
@@ -317,6 +318,11 @@ def update(
         data = self.to_model(data, "update")
         if isinstance(id_attribute, InstrumentedAttribute):
             id_attribute = cast("str", id_attribute.description)
+        if item_id is None and self.repository.get_id_attribute_value(item=data, id_attribute=id_attribute) is None:
+            msg = f"Could not identify ID attribute value.  One of the following is required: ``item_id`` or ``data.{id_attribute or self.repository.id_attribute}``"
```
peterschutt, line 322:
```suggestion
            msg = (
                "Could not identify ID attribute value.  One of the following is required: "
                f"``item_id`` or ``data.{id_attribute or self.repository.id_attribute}``"
            )
```

advanced_alchemy/service/_async.py
peterschutt: The way GH wraps these lines makes these edits look horrendous, but its just breaking the string up over two lines.

peterschutt (approved):
Thanks mate! Just a few style nits.

PR #77, opened 2023-10-25
feat: add `get_and_update` to the repository and services

cofin:
Adds `get_and_update` functions to the repository.  If the record is not found, and exception is raised. 

-------

Now that we've renamed the method to `get_or_upsert`, there have been a few cases where I have needed to temporarily disable one of the operations on the endpoint.  @JacobCoffee asked for something similar recently as well.

Rather than add a `get_and_update` method that disables the "insert" operation of the `get_or_upsert`, I am proposing we have an additional parameter (and rename an existing one) to optionally disable the create and update operations on the function.

This PR:
- renames the `upsert` parameter to `update`.
- adds a `create` parameter

Both of these default to `True` to ensure current functionality stays consistent.

### DIFF
--- a/advanced_alchemy/repository/_async.py
+++ b/advanced_alchemy/repository/_async.py
@@ -485,7 +485,8 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
         """
         return await self.get_or_upsert(
             match_fields=match_fields,
-            upsert=upsert,
+            update=upsert,
+            create=upsert,
             attribute_names=attribute_names,
             with_for_update=with_for_update,
             auto_commit=auto_commit,
@@ -497,7 +498,7 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
     async def get_or_upsert(
         self,
         match_fields: list[str] | str | None = None,
-        upsert: bool = True,
+        update: bool = True,
         create: bool = True,
         attribute_names: Iterable[str] | None = None,
         with_for_update: bool | None = None,
@@ -511,7 +512,7 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
         Args:
             match_fields: a list of keys to use to match the existing model.  When
                 empty, all fields are matched.
-            upsert: When using match_fields and actual model values differ from
+            update: When using match_fields and actual model values differ from
                 `kwargs`, perform an update operation on the model.
             create: Should a model be created.  If no model is found, an exception is raised.
             attribute_names: an iterable of attribute names to pass into the ``update``
@@ -548,7 +549,7 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
             self.check_not_found(existing)
         if not existing:
             return await self.add(self.model_type(**kwargs)), True  # pyright: ignore[reportGeneralTypeIssues]
-        if upsert:
+        if update:
             for field_name, new_field_value in kwargs.items():
                 field = getattr(existing, field_name, None)
                 if field and field != new_field_value:
--- a/advanced_alchemy/repository/_async.py
+++ b/advanced_alchemy/repository/_async.py
@@ -485,8 +485,7 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
         """
         return await self.get_or_upsert(
             match_fields=match_fields,
-            update=upsert,
-            create=upsert,
+            upsert=upsert,
             attribute_names=attribute_names,
             with_for_update=with_for_update,
             auto_commit=auto_commit,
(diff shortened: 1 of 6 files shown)


advanced_alchemy/repository/_async.py
```diff
@@ -512,6 +513,7 @@ async def get_or_upsert(
                 empty, all fields are matched.
             upsert: When using match_fields and actual model values differ from
                 `kwargs`, perform an update operation on the model.
+            create: Should a model be created.  If no model is found, an exception is raised.
```
peterschutt, line 516: OK, sorry I'm starting to get a bit confused.

We call it `get_or_upsert()` where "upsert" means to update existing or insert new. However, "get_or_" implies that if we can get an instance, then we don't upsert. But then that means that if we can get an instance, we can't upsert because we've would have been able to "get" it.

Now adding a flag to disable creation means we can disable the "sert" part of the upsert as well.
Is this a sign that we are jamming too much functionality in this method??

(Sorry if I'm missing context, but no linked issue or description on the PR.)

cofin: @JacobCoffee you want to explain your use case on this one?

cofin: @peterschutt , I think we need to change the `upsert` attribute to `perform_update` or something similar.  Then the addition of the `create` or `perform_create` operation would make more sense.

If I understood @JacobCoffee use case, he wanted to fetch and update.  But he didn't want to create an object if it doesn't exist.

cofin: I also added a bit of context for the change

JacobCoffee: My use case here is I want to get something and update it only. I made a `get_and_update` method in my work project but @cofin  (and I? cant remember) thought it would be better to just add on to the existing `get_and_upsert` with the "add if not exist" turned off. 

It maybe does feel crammed in, but I'm not sure if I mind? I'm good for whichever path we decide

peterschutt: > wanted to fetch and update

Sounds like a good method name :)

Imagine you are a future reader of this code, what is clearer:

- `repo.fetch_and_update(...)` something along these lines
- `repo.get_or_upsert(data, create=False)`.

JacobCoffee: > `repo.fetch_and_update(...)`

I like this even if it adds more methods to look up/remember

cofin: To really argument to any of this.   To be consistent with naming, I recommend we go with `get_and_update`.  We don't use the `fetch` verb anywhere else, so it may be confusing.

advanced_alchemy/repository/_async.py
```diff
@@ -560,6 +568,74 @@ async def get_or_upsert(
...
+        with_for_update: bool | None = None,
+        auto_commit: bool | None = None,
+        auto_expunge: bool | None = None,
+        auto_refresh: bool | None = None,
+        **kwargs: Any,
+    ) -> tuple[ModelT, bool]:
+        """Get instance identified by ``kwargs`` and update the model if the arguments are different.
+
+
+
```
peterschutt, line 584:
```suggestion
```

advanced_alchemy/repository/_sync.py
```diff
@@ -561,6 +569,74 @@ def get_or_upsert(
...
+        with_for_update: bool | None = None,
+        auto_commit: bool | None = None,
+        auto_expunge: bool | None = None,
+        auto_refresh: bool | None = None,
+        **kwargs: Any,
+    ) -> tuple[ModelT, bool]:
+        """Get instance identified by ``kwargs`` and update the model if the arguments are different.
+
+
+
```
peterschutt, line 585:
```suggestion
```

advanced_alchemy/service/_async.py
```diff
@@ -433,7 +433,7 @@ async def upsert_many(
 
     async def get_or_upsert(
         self,
-        match_fields: list[str] | None = None,
+        match_fields: list[str] | str | None = None,
```
peterschutt, line 436: Intended?

advanced_alchemy/service/_async.py
```diff
@@ -449,6 +449,7 @@ async def get_or_upsert(
                 empty, all fields are matched.
             upsert: When using match_fields and actual model values differ from
                 `kwargs`, perform an update operation on the model.
+            create: Should a model be created.  If no model is found, an exception is raised.
```
peterschutt, line 452: Intended?

peterschutt (approved):

advanced_alchemy/service/_async.py
cofin: Yep.  This was one of the few places I had not made that update.

advanced_alchemy/service/_async.py
cofin, line 452:
```suggestion
```

cofin:
```suggestion
```

PR #79, opened 2023-10-25
feat: `upsert` and `upsert_many` now support `match_fields`

cofin:
This PR add additional logic to the `upsert` and `upsert_many` operations so that they will now leverage `match_fields` for the `on` operation of the merge.

When leveraging multiple fields, a query to fetch data is created that uses the following pattern:
```sql
WHERE `id` in (...)
or (`name` in (...) and `other_id_column` in (..) and ...)
```

### DIFF
--- a/advanced_alchemy/repository/_async.py
+++ b/advanced_alchemy/repository/_async.py
@@ -955,7 +955,7 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
         match_filter: list[FilterTypes | ColumnElement[bool]] = [
             CollectionFilter(
                 field_name=self.id_attribute,
-                values=[getattr(datum, self.id_attribute) for datum in data] if data else None,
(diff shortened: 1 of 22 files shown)


JacobCoffee (dismissed):

advanced_alchemy/repository/_async.py
```diff
@@ -58,6 +58,7 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
     _prefer_any: bool = False
     prefer_any_dialects: tuple[str] | None = ("postgresql",)
     """List of dialects that prefer to use ``field.id = ANY(:1)`` instead of ``field.id IN (...)``."""
+    _prefer_merge: bool
```
cofin, line 61:
```suggestion
```

peterschutt (approved):

PR #80, opened 2023-10-27
feat: add `simple_asdict` and use the ASGI lifecycle to manage state on restart

cofin:


peterschutt (approved):
How good are the lifespan contexts!

PR #83, opened 2023-10-30
feat: adds a get_session function to the config

cofin:
Generally most apps end up creating some sort of context manager to help with sessions.  While the Litestar plugin handles this for us, I still find the need to get to a context manager outside of requests.  See here for an [example](https://github.com/litestar-org/litestar-fullstack/blob/main/src/app/lib/db/base.py#L103).

This PR enables quick access to a `Session`,`AsyncSession` context manager from the SQL Alchemy configuration.

-------------------
Adds a context manager for a DB session to the configuration

### DIFF
--- a/advanced_alchemy/config/asyncio.py
+++ b/advanced_alchemy/config/asyncio.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
+from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, AsyncGenerator
 
 from sqlalchemy.ext.asyncio import AsyncConnection, AsyncEngine, AsyncSession, async_sessionmaker, create_async_engine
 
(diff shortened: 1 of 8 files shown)


peterschutt:
What's the intended use-case for these?

Alc-Alc (approved):

cofin:
> What's the intended use-case for these?

To replace [this](https://github.com/litestar-org/litestar-fullstack/blob/main/src/app/lib/db/base.py#L103)

I have quite a few cases where I want to get a session quickly outside of Litestar.  This allows me to use the same config and provides a way to get a session.

That said, maybe it should be moved to the parent configuration classes instead of litestar's.

peterschutt:
> > What's the intended use-case for these?
> 
> To replace [this](https://github.com/litestar-org/litestar-fullstack/blob/main/src/app/lib/db/base.py#L103)
> 
> I have quite a few cases where I want to get a session quickly outside of Litestar. This allows me to use the same config and provides a way to get a session.

It would be nice to include this sort of context in PR descriptions. I'm not against the change at all, but I could see we were adding the functionality, not using it ourselves, and no linked issue from which to get more context. Hence the Q.

> That said, maybe it should be moved to the parent configuration classes instead of litestar's.

Sounds good, definitely not litestar specific utility :+1:

PR #84, opened 2023-10-31
fix: adds `greenlet` as a dependencies for Mac OS

cofin:
Greenlet is a required dependency for SQLAlchemy Asyncio on OSX.  This change will automatically install `greenlet` for this platform.


peterschutt (approved):
Thanks!

PR #86, opened 2023-10-31
feat: enable `flask` support

cofin:
The PR changes how we validate the `session` object.

In normal Sessions, the `bind` attribute is never empty.  However, when using scoped sessions, this is intended to be empty.

This change will call the `get_bind()` function when the bind is not set.


Alc-Alc (approved):
that was fast 🙂

PR #90, opened 2023-11-03
fix: change `upsert_many` behavior

cofin:
This PR changes the `upsert_many` logic.

Previously, it had no way of merging the identify of the existing objects back to the data being inserted.  This change will merge the identiy value on to the data to be udpated.

### DIFF
--- a/advanced_alchemy/repository/_async.py
+++ b/advanced_alchemy/repository/_async.py
@@ -526,8 +526,9 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
             When using match_fields and actual model values differ from ``kwargs``, the
             model value will be updated.
         """
-        match_fields = self._get_match_fields(match_fields=match_fields, id_attribute=None)
-        if match_fields:
+        if match_fields := self._get_match_fields(
+            match_fields=match_fields, id_attribute=None
+        ):
             match_filter = {
                 field_name: kwargs.get(field_name, None)
                 for field_name in match_fields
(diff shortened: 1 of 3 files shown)


advanced_alchemy/repository/_async.py
```diff
@@ -1016,26 +1016,40 @@ async def upsert_many(
...
+        if match_fields is None:
+            match_fields = [self.id_attribute]
+        match_filter: list[FilterTypes | ColumnElement[bool]] = []
         if match_fields:
             for field_name in match_fields:
                 field = get_instrumented_attr(self.model_type, field_name)
-                matched_values = [getattr(datum, field_name) for datum in data if datum is not None]
+                matched_values = [
+                    getattr(datum, field_name) for datum in data if getattr(datum, field_name) is not None
+                ]
```
provinzkraut, line 1027:
```suggestion
                    field_data for datum in data if (field_data := getattr(datum, field_name)) is not None
                ]
```

avoids the duplicated `getattr` call

advanced_alchemy/repository/_async.py
```diff
@@ -1016,26 +1016,40 @@ async def upsert_many(
...
+                matched_values = [getattr(datum, field_name) for datum in existing_objs if datum is not None]
+                if self._prefer_any:
+                    match_filter.append(any_(matched_values) == field)  # type: ignore[arg-type]
+                else:
+                    match_filter.append(field.in_(matched_values))
+            existing_ids = [
+                getattr(datum, self.id_attribute)
+                for datum in existing_objs
+                if getattr(datum, self.id_attribute) is not None
+            ]
```
provinzkraut, line 1049:
```suggestion
            existing_ids = [
                obj_id
                for datum in existing_objs
                if (obj_id := getattr(datum, self.id_attribute)) is not None
            ]
```

Same as above. I think this might also be worth extracting into its own utility method

provinzkraut (approved):
Looks good overall, just two small things I'd change

PR #93, opened 2023-11-07
feat: convert `None` to `is null`

cofin:
This PR allows you to use `None` in a `kwarg` filter on queries.

it will automatically convert `repo.list(field=None)` to a `field is null` query instead of `field = null`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,7 @@ dev = [
   "pyodbc>=4.0.39",
   "duckdb>=0.9.0",
   "asgi-lifespan>=2.1.0",
+  "aioodbc>=0.5.0",
 ]
 docs = [
   "sphinx>=7.1.2",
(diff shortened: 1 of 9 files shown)


guacs (approved):

cofin:
This looks like is something that's actually already handled by SQLAlchemy.  Since the original issue is not currently reproducible, setting PR to draft.

pyproject.toml
```diff
@@ -32,11 +32,11 @@ dependencies = [
 ]
 description = "Ready-to-go SQLAlchemy concoctions."
 keywords = ["sqlalchemy", "alembic", "litestar", "sanic", "fastapi", "flask"]
-license = { text = "MIT" }
+license = {text = "MIT"}
 name = "advanced_alchemy"
 readme = "README.md"
 requires-python = ">=3.8"
-version = "0.5.5"
+version = "0.5.3"
```
cofin, line 39:
```suggestion
version = "0.5.5"
```

cofin:
We'll revisit this later if needed.

PR #95, opened 2023-11-08
fix: adds the `python_type` property.  

cofin:
This implements the `python_type` property for `DateTimeUTC`.

This is used by other libraries such as `polyfactory`.  It was previously raising a `NotImplementedError` instead of returning a `datetime` type.


Alc-Alc (approved):

rseeley:
Thanks for jumping on this so quickly!

PR #11, opened 2023-09-15
feat: remove generic and abstract repositories

provinzkraut:
[//]: # "By submitting this pull request, you agree to:"
[//]: # "- follow [Jolt's Code of Conduct](https://github.com/jolt-org/.github/blob/main/CODE_OF_CONDUCT.md)"
[//]: # "- follow [PSFs's Code of Conduct](https://www.python.org/psf/conduct/)"
[//]: # "- follow [Jolt's contribution guidelines](https://github.com/jolt-org/.github/blob/main/CONTRIBUTING.md)"

### Pull Request Checklist

- [ ] New code has 100% test coverage
- [ ] (If applicable) The prose documentation has been updated to reflect the changes introduced by this PR
- [ ] (If applicable) The reference documentation has been updated to reflect the changes introduced by this PR
- [ ] Pre-Commit Checks were ran and passed
- [ ] Tests were ran and passed

### Description
[//]: # "Please describe your pull request for new release changelog purposes"

- 

### Close Issue(s)
[//]: # "Please add in issue numbers this pull request will close, if applicable"
[//]: # "Examples: Fixes #4321 or Closes #1234"

-

PR #100, opened 2023-11-26
feat: mock repository

gazorby:

### Pull Request Checklist

### Description

Add an in-memory repository intended for testing purposes.

This is a downstream implementation I've had for a few months. I'm sharing it here so we can see if it's worth merging and, if so, what it should look like.

The main goal is to have something we can use to substitute SQLAlchemy repositories when the database is not available (eg: unit testing), so the mocked interface would work where the SQLAlchemy variant does.

The limitation is that part of the repository interface that is SQLAlchemy or SQL-specific will have no effect. This includes filtering using `ColumnElement`, using parameters like, `statement`, `with_for_update`, `attribute_name` etc.

```python
mocked_repo.list(name="foo") # Return the filtered collection as expected
mocked_repo.list(MyModel.name == "foo") # Works but the collection is not actually filtered (so it's the same as calling mocked_repo.list())
mocked_repo.get("1", statement=select(MyModel).where(MyModel.name == "foo")) # The statement argument has no effects, so the collection is only filtered by the `id_attribute`
```

### Close Issue(s)

-


cofin:
Thanks for this.  I'll spend some time reviewing today.

cofin (approved):
I honestly can't add to much to this.  It looks great IMO.

Really nice work!

@provinzkraut or @peterschutt, do either of you want to weigh in before I merge this?

gazorby:
As we now have two repository implementations, can't we add back the abstract repo removed in https://github.com/jolt-org/advanced-alchemy/pull/11?

cofin:
> As we now have two repository implementations, can't we add back the abstract repo removed in #11?

We were hoping to simplify things a bit by removing it, but I think it's probably the right time to re-introduce this.

PR #112, opened 2023-12-06
fix: add Alembic template for Encrypted String

cofin:
Updates the alembic template so that it will properly generate the DDL for Encrypted types.

Without the default key, the Alembic templates will complain- even though the key is not actually needed to deploy the DDLs


JacobCoffee (approved):
lgtm

PR #113, opened 2023-12-07
fix: adds `MISSING` placeholder

cofin:
This PR corrects an issue when the existing value of a field is None and you wish to update it.

### DIFF
--- a/advanced_alchemy/repository/_async.py
+++ b/advanced_alchemy/repository/_async.py
@@ -548,7 +548,7 @@ class SQLAlchemyAsyncRepository(Generic[ModelT]):
         if upsert:
             for field_name, new_field_value in kwargs.items():
                 field = getattr(existing, field_name, MISSING)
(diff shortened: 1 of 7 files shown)


advanced_alchemy/repository/_async.py
```diff
@@ -547,7 +547,7 @@ async def get_or_upsert(
             )
         if upsert:
             for field_name, new_field_value in kwargs.items():
-                field = getattr(existing, field_name, None)
+                field = getattr(existing, field_name, MISSING)
                 if field and field != new_field_value:
```
provinzkraut, line 551:
```suggestion
                field = getattr(existing, field_name, MISSING)
                if field is not MISSING and field != new_field_value:
```

Otherwise the check will never be `False`

provinzkraut (changes requested):

provinzkraut (approved):

PR #138, opened 2024-03-09
feat: `UniqueMixin` that instantiates objects ensuring uniqueness on some field(s)

Alc-Alc:
This PR introduces "Unique Objects", a slight modified implementation derived from https://github.com/sqlalchemy/sqlalchemy/wiki/UniqueObject

### Pull Request Checklist

- [ ] Pre-Commit Checks were ran and passed

### Description

-  Includes `UniqueMixin` featuring two methods: `as_unique_sync` and `as_unique_async`. These can be used to instantiate objects while ensuring uniqueness on some field(s).

### Close Issue(s)

-

### DIFF
--- a/advanced_alchemy/exceptions.py
+++ b/advanced_alchemy/exceptions.py
@@ -84,3 +84,7 @@ class IntegrityError(RepositoryError):
 
 class NotFoundError(RepositoryError):
     """An identity does not exist."""
+
+
+class MultipleResultsFoundError(AdvancedAlchemyError):
+    """A single database result was required but more than one were found."""
--- a/advanced_alchemy/exceptions.py
+++ b/advanced_alchemy/exceptions.py
@@ -96,8 +96,8 @@ class MultipleResultsFoundError(AdvancedAlchemyError):
 
 @contextmanager
 def wrap_sqlalchemy_exception() -> Any:
-    """Do something within context to raise a `RepositoryError` chained
-    from an original `SQLAlchemyError`.
+    """Do something within context to raise a ``RepositoryError`` chained
+    from an original ``SQLAlchemyError``.
 
         >>> try:
         ...     with wrap_sqlalchemy_exception():
--- a/advanced_alchemy/exceptions.py
+++ b/advanced_alchemy/exceptions.py
@@ -1,6 +1,10 @@
 from __future__ import annotations
 
+from contextlib import contextmanager
 from typing import Any
+
+from sqlalchemy.exc import IntegrityError as SQLAlchemyIntegrityError
+from sqlalchemy.exc import MultipleResultsFound, SQLAlchemyError
 
 from advanced_alchemy.utils.deprecation import deprecated
 
@@ -88,3 +92,30 @@ class NotFoundError(RepositoryError):
 
 class MultipleResultsFoundError(AdvancedAlchemyError):
     """A single database result was required but more than one were found."""
+
+
+@contextmanager
+def wrap_sqlalchemy_exception() -> Any:
+    """Do something within context to raise a `RepositoryError` chained
+    from an original `SQLAlchemyError`.
+
+        >>> try:
+        ...     with wrap_sqlalchemy_exception():
+        ...         raise SQLAlchemyError("Original Exception")
+        ... except RepositoryError as exc:
+        ...     print(f"caught repository exception from {type(exc.__context__)}")
+        ...
+        caught repository exception from <class 'sqlalchemy.exc.SQLAlchemyError'>
+    """
+    try:
+        yield
+    except MultipleResultsFound as e:
+        msg = "Multiple rows matched the specified key"
+        raise MultipleResultsFoundError(msg) from e
+    except SQLAlchemyIntegrityError as exc:
+        raise IntegrityError from exc
+    except SQLAlchemyError as exc:
+        msg = f"An exception occurred: {exc}"
+        raise RepositoryError(msg) from exc
+    except AttributeError as exc:
+        raise RepositoryError from exc
--- a/advanced_alchemy/mixins/__init__.py
+++ b/advanced_alchemy/mixins/__init__.py
@@ -1,0 +1,3 @@
+from .unique import UniqueMixin
+
+__all__ = ("UniqueMixin",)
--- a/advanced_alchemy/mixins/unique.py
+++ b/advanced_alchemy/mixins/unique.py
@@ -15,6 +15,8 @@ if TYPE_CHECKING:
 
 
 class UniqueMixin:
+    """Mixin for instantiating objects while ensuring uniqueness on some field(s)."""
+
     @classmethod
     async def as_unique_async(
         cls,
@@ -22,6 +24,18 @@ class UniqueMixin:
         *args: Any,
         **kwargs: Any,
     ) -> Self:
+        """Instantiate and return a unique object within the provided session based on the given arguments.
+
+        If an object with the same unique identifier already exists in the session, it is returned from the cache.
+
+        Args:
+            session (AsyncSession | async_scoped_session[AsyncSession]): SQLAlchemy async session
+            *args (Any): Columns belonging to a table
+            **kwargs (Any): Columns belonging to a table
+
+        Returns:
+            Self: The unique object instance.
+        """
         key = cls, cls.unique_hash(*args, **kwargs)
         cache: dict[tuple[type[Self], Hashable], Self] | None = getattr(session, "_unique_cache", None)
         if cache is None:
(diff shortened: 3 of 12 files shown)


advanced_alchemy/mixins/unique.py
```diff
@@ -0,0 +1,117 @@
...
+    from collections.abc import Hashable
+
+    from sqlalchemy.ext.asyncio import AsyncSession
+    from sqlalchemy.ext.asyncio.scoping import async_scoped_session
+    from sqlalchemy.orm import Session
+    from sqlalchemy.orm.scoping import scoped_session
+    from typing_extensions import Self
+
+
+class UniqueMixin:
```
Alc-Alc, line 17: I ended up not writing two different mixins (sync and async) as I felt doing so would make the user "lock in" to "sync only" or "async only" flows. If one wants to support both(sync and async), one would intuitively try to inherit from both the mixins and I was not sure if that would play well with the "inheritance chain".

So I added two `as_unique` methods that can coexist together.

advanced_alchemy/mixins/unique.py
```diff
@@ -0,0 +1,117 @@
...
+                session.add(obj := cls(*args, **kwargs))
+        cache[key] = obj
+        return obj
+
+    @classmethod
+    def unique_hash(cls, *arg: Any, **kw: Any) -> Hashable:  # noqa: ARG003
+        """Generate a unique key based on the provided arguments.
+
+        This method should be implemented in the subclass.
+
```
peterschutt, line 94: We should document that `arg` and `kw` are the values passed into `as_unique_{a|s}ync`.. also I think they should be named `args` and `kwargs` to be consistent with the other methods.

advanced_alchemy/mixins/unique.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        msg = "Implement this in subclass"
+        raise NotImplementedError(msg)
+
+    @classmethod
+    def unique_filter(cls, *arg: Any, **kw: Any) -> ColumnElement[bool]:  # noqa: ARG003
+        """Generate a filter condition for ensuring uniqueness.
+
+        This method should be implemented in the subclass.
+
```
peterschutt, line 109: Same

advanced_alchemy/mixins/unique.py
peterschutt: Yeh, I agree with this. ORM types are otherwise totally interchangeable between sync/async paradigms so it would be a shame to introduce some division down those lines here.

advanced_alchemy/mixins/unique.py
```diff
@@ -0,0 +1,117 @@
...
+        Args:
+            session (AsyncSession | async_scoped_session[AsyncSession]): SQLAlchemy async session
+            *args (Any): Columns belonging to a table
+            **kwargs (Any): Columns belonging to a table
+
+        Returns:
+            Self: The unique object instance.
+        """
+        key = cls, cls.unique_hash(*args, **kwargs)
```
peterschutt, line 39: I'd move this declaration to right before `key` is actually used, just above `if obj := cache.get(key):`

advanced_alchemy/mixins/unique.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Args:
+            session (Session | scoped_session[Session]): SQLAlchemy sync session
+            *args (Any): Columns belonging to a table
+            **kwargs (Any): Columns belonging to a table
+
+        Returns:
+            Self: The unique object instance.
+        """
+        key = cls, cls.unique_hash(*args, **kwargs)
```
peterschutt, line 74: same

advanced_alchemy/mixins/unique.py
```diff
@@ -0,0 +1,117 @@
...
+        cache: dict[tuple[type[Self], Hashable], Self] | None = getattr(session, "_unique_cache", None)
+        if cache is None:
+            cache = {}
+            setattr(session, "_unique_cache", cache)
+        if obj := cache.get(key):
+            return obj
+
+        with session.no_autoflush:
+            statement = select(cls).where(cls.unique_filter(*args, **kwargs)).limit(1)
```
peterschutt, line 48: Instead of `.limit(1)` DYT we should use `.limit(2)` and then use `.one_or_none()` instead of `first()`? This way, if whatever clause is provided by `cls.unique_filter()` doesn't actually limit the result set to a single (or no) row it will be an error as opposed to passing silently.

peterschutt (changes requested):
LGTM @Alc-Alc - my comments mostly nits.

Is the idea of these to reduce needing to pass instances around?

Should we have something added to docs?

Oh, and maybe also a reference to the sqlalchemy wiki that the pattern is derived from in the docstring of the mixin.

advanced_alchemy/mixins/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .unique import UniqueMixin
+
+__all__ = ("UniqueMixin",)
```
cofin, line 3: Can we re-export `mixins` at the top level and at `advanced_alchemy.extensions.litestar`

The reason being is purely to minimize imports required for using in Litestar

In a litestar app:
```python
from advanced_alchemy.extensions.litestar import types, mixins, service, repository
...
```

cofin (approved):
LGTM.  I agree with all of @peterschutt's suggested edits as well.

advanced_alchemy/mixins/__init__.py
cofin: On second thought, we can clean this up in a followup pr.  There are a few other organization changes to make, I think.

tests/integration/test_unique_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+    @classmethod
+    def unique_hash(cls, col_1: int, col_2: int, col_3: str) -> Hashable:
+        return (col_1, col_3)
+
+    @classmethod
+    def unique_filter(cls, col_1: int, col_2: int, col_3: str) -> ColumnElement[bool]:
+        return (cls.col_1 == col_1) & (cls.col_3 == col_3)
+
+
+class BigIntModelWithMaybeUniqueValue(UniqueMixin, BigIntBase):
```
Alc-Alc, line 45: I have no idea what to name this model 😞 open to better names

JacobCoffee: seems okay for tests :)

JacobCoffee (approved):
good cleanup there at the end

advanced_alchemy/mixins/unique.py
```diff
@@ -0,0 +1,175 @@
...
+
+class UniqueMixin:
+    """Mixin for instantiating objects while ensuring uniqueness on some field(s).
+
+    This is a slightly modified implementation derived from https://github.com/sqlalchemy/sqlalchemy/wiki/UniqueObject
+    """
+
+    @classmethod
+    @contextmanager
+    def _uniquify(
```
cofin, line 40:
```suggestion
    def _no_autoflush(
```

We may have other exceptions we want to check for in the future and that would let us re-use this.

cofin (approved):

advanced_alchemy/exceptions.py
```diff
@@ -84,3 +88,34 @@ class IntegrityError(RepositoryError):
...
+
+
+class MultipleResultsFoundError(AdvancedAlchemyError):
+    """A single database result was required but more than one were found."""
+
+
+@contextmanager
+def wrap_sqlalchemy_exception() -> Any:
+    """Do something within context to raise a `RepositoryError` chained
+    from an original `SQLAlchemyError`.
```
JacobCoffee, line 100:
```suggestion
    """Do something within context to raise a ``RepositoryError`` chained
    from an original ``SQLAlchemyError``.
```

advanced_alchemy/mixins/unique.py
```diff
@@ -0,0 +1,161 @@
...
+        cls,
+        session: Session | scoped_session[Session],
+        *args: Any,
+        **kwargs: Any,
+    ) -> Self:
+        """Instantiate and return a unique object within the provided session based on the given arguments.
+
+        If an object with the same unique identifier already exists in the session, it is returned from the cache.
+
+
```
JacobCoffee, line 98:
```suggestion

```

advanced_alchemy/exceptions.py
```diff
@@ -84,3 +88,34 @@ class IntegrityError(RepositoryError):
...
+
+        >>> try:
+        ...     with wrap_sqlalchemy_exception():
+        ...         raise SQLAlchemyError("Original Exception")
+        ... except RepositoryError as exc:
+        ...     print(f"caught repository exception from {type(exc.__context__)}")
+        ...
+        caught repository exception from <class 'sqlalchemy.exc.SQLAlchemyError'>
```
JacobCoffee, line 108: is this valid? does sphinx expect `.. code-block::` directive instead?

JacobCoffee (approved):

advanced_alchemy/exceptions.py
cofin: This hasn't been modified from what it was previously.

JacobCoffee: it works!

cofin (approved):

PR #141, opened 2024-03-16
fix: accept `None` in `model_from_dict` to allow setting model fields…

rseeley:
… to `NULL`

This change allows for sending a dict with `None` as values; the model will then be updated to `NULL` for those fields. The existing handling in `model_from_dict` will drop dict keys if they're set to None.

### Pull Request Checklist

### Description

- Allows for sending a dict with `None` as values to `model_from_dict`; the model will then be updated to `NULL` for those fields. This changes the existing behavior of `model_from_dict`, which will drop dict keys if they're set to `None`, causing the record's field to be unchanged.


tests/integration/test_repository.py
```diff
@@ -1921,6 +1921,26 @@ async def test_service_update_method_no_item_id(author_service: AuthorService, f
...
 
 
+async def test_service_update_method_data_is_dict(author_service: AuthorService, first_author_id: Any) -> None:
+    new_date = datetime.date(datetime.now())
+    updated_obj = await maybe_async(
+        author_service.update(item_id=first_author_id, data={"dob": new_date}),
+    )
+    assert updated_obj.dob == new_date
+    # ensure the other fields are not affected
+    assert updated_obj.name == "Agatha Christie"
```
rseeley, line 1931: This, and the similar assertion in `test_service_update_method_data_is_dict_with_none_value`, fail for the `sync-uuid-mock_sync_engine` and `async-uuid-mock_async_engine`, saying the `name` value is `None` here. This is the case for the existing version of `model_from_dict` as well as this newer version, so it's not a regression. I ran out of time so I couldn't dig further into this for the time being, so I made this a draft PR. Wondering if anyone else has thoughts about this, though.

cemrehancavdar:
Today we needed that! If you need an hand you can ping us :) @ysnbyzli and me

rseeley:
> Today we needed that! If you need a hand you can ping us :) @ysnbyzli and me

Will do! Sorry for the delay on this. I've been super sidetracked. I'm hoping I can get to it some time before the end of the weekend.

cofin:
> > Today we needed that! If you need a hand you can ping us :) @ysnbyzli and me
> 
> Will do! Sorry for the delay on this. I've been super sidetracked. I'm hoping I can get to it some time before the end of the weekend.

I've actually got a working implementation that is non-breaking.

cofin:
To get the new behavior, you pass `exclude_unset` as `False` into the any of the service calls that perform an update.

cofin (approved):

rseeley:
> To get the new behavior, you pass `exclude_unset` as `False` into the any of the service calls that perform an update.

I haven't been able to test this myself yet as I have to run till later tonight, but just curious, does this change require you to explicitly set `False` if you are sending one or more values as `None` for it to work? What happens when you pass one field as `None` but others that aren't `None` in either scenario (`True` and `False`)? I do plan to do my own testing, but wanted to throw it out there. The "unset" nomenclature is a bit confusing to me since explicitly passing `None`, to me, is "set", it's just set to a null value. Is there perhaps a clean way to handle this without having to pass a flag for better DX?

cofin:
> > To get the new behavior, you pass `exclude_unset` as `False` into the any of the service calls that perform an update.
> 
> I haven't been able to test this myself yet as I have to run till later tonight, but just curious, does this change require you to explicitly set `False` if you are sending one or more values as `None` for it to work? What happens when you pass one field as `None` but others that aren't `None` in either scenario (`True` and `False`)? I do plan to do my own testing, but wanted to throw it out there. The "unset" nomenclature is a bit confusing to me since explicitly passing `None`, to me, is "set", it's just set to a null value. Is there perhaps a clean way to handle this without having to pass a flag for better DX?

Hold off on testing.  I don't think the change I've pushed are what you want.  Give me a bit to test a few things out, and I'll push some additional changes shortly.

cofin:
I was able to get the original issue with the mock repository resolved bit hit a new issue with the `test_lazy_load` in the repository mocks.  This looks to be unrelated, so for now, I set the lazy load test to skip the mock repository.  I'll address that issues in a followup PR.

cofin (approved):

rseeley:
> I was able to get the original issue with the mock repository resolved bit hit a new issue with the `test_lazy_load` in the repository mocks.  This looks to be unrelated, so for now, I set the lazy load test to skip the mock repository.  I'll address that issues in a followup PR.

Beautiful, thanks for your help, @cofin!

Issue #147, opened 2024-03-26
Bug: TestClient used in pytest fails with `KeyError: 'session_maker_class'`


sherbang:
### Description

I've developed a failing test by writing a test that tries to test the litestar example in this repo:  https://github.com/jolt-org/advanced-alchemy/compare/main...sherbang:advanced-alchemy:litestar_test

### URL to code causing the issue

https://github.com/sherbang/advanced-alchemy/tree/litestar_test

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

"In the format of: `![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)`"

### Logs

```bash
/home/sherbang/devel/advanced-alchemy/tests/examples/test_litestar.py::test_create failed: test_client = <litestar.testing.client.sync_client.TestClient object at 0x7582d98e9c50>

    async def test_create(test_client: TestClient[Litestar]) -> None:
        author = AuthorCreate(name="foo")
    
        response = test_client.post(
            "/authors",
            json=author.model_dump(mode="json"),
        )
>       assert response.status_code == 200, response.text
E       AssertionError: Traceback (most recent call last):
E           File "/home/sherbang/devel/advanced-alchemy/.venv/lib/python3.12/site-packages/litestar/middleware/exceptions/middleware.py", line 218, in __call__
E             await self.app(scope, receive, send)
E           File "/home/sherbang/devel/advanced-alchemy/.venv/lib/python3.12/site-packages/litestar/routes/http.py", line 82, in handle
E             response = await self._get_response_for_request(
E                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E           File "/home/sherbang/devel/advanced-alchemy/.venv/lib/python3.12/site-packages/litestar/routes/http.py", line 134, in _get_response_for_request
E             return await self._call_handler_function(
E                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E           File "/home/sherbang/devel/advanced-alchemy/.venv/lib/python3.12/site-packages/litestar/routes/http.py", line 154, in _call_handler_function
E             response_data, cleanup_group = await self._get_response_data(
E                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #148, opened 2024-03-26
chore(tests): Add basic test for `litestar` example

sherbang:

### Pull Request Checklist

- [ x ] Pre-Commit Checks were ran and passed

### Description

- Add test for litestar example

This test was developed to test, and is currently failing due to #147.  It seems like it would also be useful in the future to make sure that the documented example is kept up to date.

### Close Issue(s)

-

### DIFF
--- a/examples/litestar.py
+++ b/examples/litestar.py
@@ -1,79 +1,66 @@
 from __future__ import annotations
 
 from datetime import date, datetime
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, List, Union
 from uuid import UUID
 
 from litestar import Litestar
 from litestar.controller import Controller
 from litestar.di import Provide
+from litestar.exceptions import NotFoundException as LiteStarNotFoundException
 from litestar.handlers.http_handlers.decorators import delete, get, patch, post
 from litestar.pagination import OffsetPagination
 from litestar.params import Parameter
-from pydantic import BaseModel as _BaseModel
-from pydantic import TypeAdapter
 from sqlalchemy import ForeignKey, select
 from sqlalchemy.orm import Mapped, mapped_column, relationship, selectinload
+from typing_extensions import Annotated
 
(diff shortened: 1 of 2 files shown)


sherbang:
The litestar example was quite broken.  The pydantic models were not able to handle the UUID type from `uuid_utils`.

I noticed a comment in here about using pydantic models to not add confusion with DTOs, but I think that's only true for someone who is very proficient with Pydantic.  There was quite a lot of code that was just converting between model types.  To me, this code seems simpler, and DTOs are a good concept to learn if working with Litestar.  However, I've pulled out a lot of hair in the past getting Pydantic to properly handle custom types in the past, so adding that would definitely push the Pydantic version over the edge in terms of being more complex.

I'll address the python 3.8 type annotation issue tomorrow.  I expect I'll take the option of limiting this test to only run on main, I just need to figure out how to do that.

examples/litestar.py
```diff
@@ -196,24 +189,26 @@ async def delete_author(
...
+            connection_string="sqlite+aiosqlite:///test.sqlite",
+            session_config=session_config,
+        )  # Create 'db_session' dependency.
+
+    sqlalchemy_plugin = SQLAlchemyPlugin(config=sqlalchemy_config)
+
+    async def on_startup() -> None:
+        """Initializes the database."""
+        async with sqlalchemy_config.get_engine().begin() as conn:
+            await conn.run_sync(UUIDBase.metadata.create_all)
```
cofin, line 206: This can be removed and the `create_all` in the config can be used.

sherbang: Fixed.

I also noticed that the monkeypatch I had in conftest.py wasn't required (just moving the import into the `app()` fixture was sufficient to prevent the problems with the other tests) so I removed conftest.py.

cofin:
Sorry for the long review cycle on this.  If you'd like to fix these, let me know and we can re-open this.  There's been a number of changes that may need a second look anyway.

PR #151, opened 2024-03-29
feat: multi-database configurations

cofin:
Allows the Litestar plugin to be configured with multiple databases

### DIFF
--- a/advanced_alchemy/extensions/litestar/plugins/init/plugin.py
+++ b/advanced_alchemy/extensions/litestar/plugins/init/plugin.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import contextlib
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 from uuid import UUID
 
 from litestar.di import Provide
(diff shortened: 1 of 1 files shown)


advanced_alchemy/extensions/litestar/plugins/init/config/asyncio.py
```diff
@@ -39,29 +39,47 @@
 )
 
 
-async def default_before_send_handler(message: Message, scope: Scope) -> None:
-    """Handle closing and cleaning up sessions before sending.
-
+def default_handler_maker(
+    session_scope_key: str = SESSION_SCOPE_KEY,
```
provinzkraut, line 43: Does this need the default value? Is there some use case where we call this and don't have access to the config?

provinzkraut (approved):
Looks good!

One idea though: Since the session scope key is only used internally, how about, instead requiring the user to define one for multiple configs, we simply keep track of all the ones we have created (e.g. via a class var on the base config) and then ensure they are unique? That would make it a bit more "it just works"

advanced_alchemy/extensions/litestar/plugins/init/config/asyncio.py
cofin: Nope, it doesn't, but if I move it to the end where it should be, it'll need it.

The main reason this is there is so that we can create the `default_before_send_handler` just below this declaration.

Alc-Alc (approved):

Alc-Alc (approved):

PR #152, opened 2024-03-29
chore: prevent `create_all` failures from stopping application startup

cofin:
This PR adds an exception handler around the `create_all` startup.

Without this, this setting is only useful the first time you are deploying an environment.  Future startup will raise an exception if the tables already exist.


advanced_alchemy/extensions/litestar/cli.py
```diff
@@ -60,6 +60,8 @@ def show_database_revision(app: Litestar, verbose: bool) -> None:
 )
 def downgrade_database(app: Litestar, revision: str, sql: bool, tag: str | None, no_prompt: bool) -> None:
     """Downgrade the database to the latest revision."""
+    from advanced_alchemy.extensions.litestar.alembic import AlembicCommands
```
cofin, line 63: I've also moved these here in an effort to reduce the amount of classes that are imported.

provinzkraut:
> We actually already have [this](https://github.com/jolt-org/advanced-alchemy/blob/main/advanced_alchemy/config/common.py#L120). By default, this setting is disabled.

Ha :upsi

PR #166, opened 2024-04-30
fix: fixed behavior for `updated_at` field when updating the model

SergeyKomVl:
## Description

- fixed behavior for `updated_at` field when updating the model

  When the model is created, it is not necessary to fill in the 'updated_at` field.
  But when updating the model, it needs to be updated automatically.

- hotfix

  Returned the default value for the field.

### DIFF
--- a/advanced_alchemy/base.py
+++ b/advanced_alchemy/base.py
@@ -193,6 +193,7 @@ class AuditColumns:
     """Date/time of instance creation."""
     updated_at: Mapped[datetime] = mapped_column(
         DateTimeUTC(timezone=True),
+        default=lambda: datetime.now(timezone.utc),
         onupdate=lambda: datetime.now(timezone.utc),
     )
     """Date/time of instance last update."""
--- a/advanced_alchemy/base.py
+++ b/advanced_alchemy/base.py
@@ -193,7 +193,7 @@ class AuditColumns:
     """Date/time of instance creation."""
     updated_at: Mapped[datetime] = mapped_column(
         DateTimeUTC(timezone=True),
-        default=lambda: datetime.now(timezone.utc),
+        onupdate=lambda: datetime.now(timezone.utc),
     )
     """Date/time of instance last update."""
 


cofin:
In this case, the `updated_at` is generally populated by the model listener added [here](https://github.com/litestar-org/advanced-alchemy/blob/main/advanced_alchemy/_listeners.py#L12).  While this may not be needed, it was originally done because it supports polymorphic relationships.  I would be open to changing this if there's something that maintains that functionality.

Re: the initial population, I don't disagree that it should not be populated initially, but I often found that it was easier to work with an object you know is not null.  

Let's see how the tests suite executes and we can see if anything different is required.

SergeyKomVl:
@cofin 
I fixed it, thanks.

I didn't notice that this field is required in the database.

As for your explanation, I agree: it should be more convenient with a non-null value.

cofin:
@SergeyKomVl quick question, are you using advanced alchemy in Litestar or on it's own?  The reason I ask is that this should be taken care of by the listener that gets attached. So, in theory, you don't even need this configuration added to the model. 

However, it occurred to me that, depending on how you are using the library, you may not invoke a configuration object.

Take a look at the test [cases here](https://github.com/litestar-org/advanced-alchemy/blob/main/tests/integration/test_repository.py#L1180-L1265) and let me know if you still need this change.

cofin (approved):

SergeyKomVl:
> @SergeyKomVl quick question, are you using advanced alchemy in Litestar or on it's own? The reason I ask is that this should be taken care of by the listener that gets attached. So, in theory, you don't even need this configuration added to the model.
> 
> However, it occurred to me that, depending on how you are using the library, you may not invoke a configuration object.
> 
> Take a look at the test [cases here](https://github.com/litestar-org/advanced-alchemy/blob/main/tests/integration/test_repository.py#L1180-L1265) and let me know if you still need this change.


@cofin 
I just started learning litestar.
Thanks for the links to the cases!

PR #167, opened 2024-05-02
fix: specify type variable of GUID

wer153:
## Description

- Specifying type variable would help static analyzers do their jobs.


cofin (approved):

cofin:
Thanks for this.  I think the failure here is a result of the move into the Litestar org.  We'll get this merged soon.

PR #183, opened 2024-05-03
chore: fix code coverage and sonar

cofin:
## Description

-


CODEOWNERS
JacobCoffee, line 1: I think we should keep this its what notifies us a PR is ready

cofin: There is a second one in the `.github` folder.

cofin: I removed this one so that it mirrors what Litestar has.

JacobCoffee: oh im stupid, when you said `.github` i thought you meant https://github.com/litestar-org/.github

JacobCoffee (approved):

PR #192, opened 2024-05-12
feat: relationship loader api

cofin:
## Description

-


tests/integration/test_loader_and_execution_options.py
```diff
@@ -0,0 +1,171 @@
...
+
+        country: Mapped[UUIDCountry] = relationship(uselist=False, back_populates="states", lazy="raise")
+
+    class USStateRepository(SQLAlchemySyncRepository[UUIDState]):
+        model_type = UUIDState
+
+    class CountryRepository(SQLAlchemySyncRepository[UUIDCountry]):
+        model_type = UUIDCountry
+
+    engine = create_engine("sqlite:///:memory:", future=True, echo=True)
```
Alc-Alc, line 31:
```suggestion
    engine = create_engine("sqlite:///:memory:", echo=True)
```

tests/integration/test_loader_and_execution_options.py
```diff
@@ -0,0 +1,171 @@
...
+
+        country: Mapped[BigIntCountry] = relationship(uselist=False, back_populates="states", lazy="raise")
+
+    class USStateRepository(SQLAlchemyAsyncRepository[BigIntState]):
+        model_type = BigIntState
+
+    class CountryRepository(SQLAlchemyAsyncRepository[BigIntCountry]):
+        model_type = BigIntCountry
+
+    engine = create_async_engine("sqlite+aiosqlite:///:memory:", future=True, echo=True)
```
Alc-Alc, line 110:
```suggestion
    engine = create_async_engine("sqlite+aiosqlite:///:memory:", echo=True)
```

tests/integration/test_loader_and_execution_options.py
```diff
@@ -0,0 +1,171 @@
...
+        california = BigIntState(name="California", country=usa)
+        oregon = BigIntState(name="Oregon", country=usa)
+        ile_de_france = BigIntState(name="Île-de-France", country=france)
+
+        repo = USStateRepository(session=db_session)
+        await repo.add(california)
+        await repo.add(oregon)
+        await repo.add(ile_de_france)
+        await db_session.commit()
+        del repo
```
Alc-Alc, line 131: I am not really sure here so feel free to ignore this comment, As the names for the repos are different, do we need this `del`? If I see a `del` I assume it either runs some clean up action or the name is reused, I don't think either is the case here.

Alc-Alc (approved):
Just some nits, those aside, this PR is *loaded*

Alc-Alc (approved):

advanced_alchemy/base.py
```diff
@@ -92,26 +95,26 @@ def merge_table_arguments(
...
             if isinstance(arg_to_merge, tuple):
                 last_positional_arg = arg_to_merge[-1]
                 args.extend(arg_to_merge[:-1])
                 if isinstance(last_positional_arg, dict):
-                    kwargs.update(last_positional_arg)
+                    kwargs.update(last_positional_arg)  # pyright: ignore[reportUnknownArgumentType]
                 else:
                     args.append(last_positional_arg)
-            elif isinstance(arg_to_merge, dict):
+            else:
```
peterschutt, line 117: I believe this can only be a `dict` at this point, if `arg_to_merge` can be `None`, a `tuple` or a `dict`, then `dict` is all that's left.

advanced_alchemy/base.py
```diff
@@ -126,8 +129,8 @@ class ModelProtocol(Protocol):
     """The base SQLAlchemy model protocol."""
 
     __table__: FromClause
-    __mapper__: Mapper
-    __name__: ClassVar[str]
+    __mapper__: Mapper[Any]
```
peterschutt, line 132: In the SQLAlchemy version of this protocol, it is generic on a single type argument, which is used to type this mapper arg: https://github.com/sqlalchemy/sqlalchemy/blob/e69ca167fa0dcd33ef8c7dccbccf127c8665472e/lib/sqlalchemy/orm/decl_base.py#L90

I _think_ typing it as `Mapper[Any]` is OK for our purposes..

advanced_alchemy/base.py
```diff
@@ -272,13 +276,13 @@ def __table_args__(cls) -> tuple | dict:
 
 
 def create_registry(
-    custom_annotation_map: dict[type, type[TypeEngine[Any]] | TypeEngine[Any]] | None = None,
+    custom_annotation_map: dict[Any, type[TypeEngine[Any]] | TypeEngine[Any]] | None = None,
```
peterschutt, line 279: SQLAlchemy type this as `Any` also: https://github.com/sqlalchemy/sqlalchemy/blob/e69ca167fa0dcd33ef8c7dccbccf127c8665472e/lib/sqlalchemy/orm/decl_api.py#L101

peterschutt (commented):
Just some comments on the things I changed in `base.py`

advanced_alchemy/repository/_async.py
```diff
@@ -53,6 +56,7 @@ class SQLAlchemyAsyncRepository(FilterableRepository[ModelT]):
 
     This is useful for certain SQLAlchemy uses cases such as applying ``contains_eager`` to a query containing a one-to-many relationship
     """
+    default_options: LoadSpec
```
provinzkraut, line 59:
```suggestion
```
This does not appear to be used anywhere

advanced_alchemy/repository/_async.py
```diff
@@ -148,12 +172,34 @@ def check_not_found(item_or_none: ModelT | None) -> ModelT:
...
+            default_options_have_wildcards=self._loader_options_have_wildcards,
+        )
+
+    def _reset_execution_options(self) -> None:
+        self._execution_options = self._default_execution_options
+
+    def _set_execution_options(self, execution_options: dict[str, Any] | None = None) -> None:
+        if execution_options is not None:
+            self._execution_options = execution_options
+
```
provinzkraut, line 194: This API, or rather its usage, is partially broken. For async usage it's fine as it's being used currently, but the sync one is not thread safe. Because this relies on attributes being mutated on an instance, and these mutations aren't guarded in any way, when this is being done in 2 threads, thread 1 might set the options, hand of to thread 2, which overwrites the options with its own and hand back the control, at which point thread 1 attempts to perform the operations with the options set by thread 2.

The same scenario could happen with async context switching if multiple coroutines make use of the options within a single function call after having called e.g. `_set_loader_options`.

This could either be fixed by using contextvars or, preferably, by simply passing in the options to the methods that might want to use them.

provinzkraut (approved):

PR #198, opened 2024-05-22
fix: `to_schema` inocrrectly checks for `ModelMetaclass` instead of `BaseModel`

cofin:
## Description

-

PR #199, opened 2024-05-22
fix(typing): ResultConverter.to_schema

provinzkraut:
Fix the typing overloads of ResultConverter.to_schema


provinzkraut:
Implemented in #198

PR #204, opened 2024-05-28
fix: set `options` and `execution_options` before calling `lambda_stmt`

cofin:
## Description

-

### DIFF
--- a/tests/integration/test_loader_and_execution_options.py
+++ b/tests/integration/test_loader_and_execution_options.py
@@ -49,8 +49,14 @@ def test_loader() -> None:
         repo.add(oregon)
         repo.add(ile_de_france)
         db_session.commit()
-
+        db_session.expire_all()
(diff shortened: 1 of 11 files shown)


tests/integration/test_loader_and_execution_options.py
```diff
@@ -81,6 +74,12 @@ class CountryRepository(SQLAlchemySyncRepository[UUIDCountry]):
         usa_country_3 = star_country_repo.get_one(name="United States of America")
         assert len(usa_country_3.states) == 2
 
+        si1_country_repo = CountryRepository(session=db_session)
+        usa_country_1 = si1_country_repo.get_one(
+            name="United States of America",
+            load=[noload(UUIDCountry.states)],
+        )
+        assert len(usa_country_1.states) == 0
```
cofin, line 82: If you move this to the first test here, then this block will pass but all the others fail.

PR #220, opened 2024-06-24
feat: add a `to_schema` parameter to services

cofin:
## Description

-


Alc-Alc (approved):
1k lines of code change to save one line for end user :fire:

Alc-Alc (approved):

PR #229, opened 2024-07-07
feat: optimize type adapter call & multi-config fixes

cofin:
## Description

-

### DIFF
--- a/advanced_alchemy/service/typing.py
+++ b/advanced_alchemy/service/typing.py
@@ -6,6 +6,7 @@ should be a SQLAlchemy model.
 
 from __future__ import annotations
 
+from functools import lru_cache
 from typing import (
     Any,
     ClassVar,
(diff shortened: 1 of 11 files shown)


advanced_alchemy/service/typing.py
```diff
@@ -20,18 +21,29 @@
...
+T = TypeVar("T")  # pragma: nocover
 try:
     from pydantic import BaseModel  # pyright: ignore[reportAssignmentType]
+    from pydantic.config import ConfigDict  # pyright: ignore[reportAssignmentType]
     from pydantic.type_adapter import TypeAdapter  # pyright: ignore[reportUnusedImport, reportAssignmentType]
 
     PYDANTIC_INSTALLED: Final[bool] = True
 except ImportError:  # pragma: nocover
 
+    class ConfigDict:  # type: ignore[no-redef] # pragma: nocover
```
JacobCoffee, line 38: could you explain for my benefit the need for these?

cofin: It's to make sure the type hinting is happy in the `_util.py` and other places within the service layer.  Since it's designed to work with or without Pydantic and Msgspec, it was much easier to centralize all of that conditional logic here than have it spread over several files.

cofin: We actually don't need `ConfigDict` though.  It was or a tweak I removed.

Alc-Alc (approved):

PR #235, opened 2024-07-17
feat: `DTOData` support in services

cofin:
## Description

This PR:
- correct an issue with the default order by contructor method
- Adds support for `DTOData` in the service layer.  It will auto-convert these types to instances.


cofin:
closes #239

Alc-Alc (approved):

PR #241, opened 2024-08-01
fix: added if statement to prevent ``script_location`` from being overwritten

chaush-server:
## Description

- Fix overwritten `alembic.ini` `script_location` by default value in `self.sqlalchemy_config.alembic_config.script_location`

### DIFF
--- a/advanced_alchemy/alembic/commands.py
+++ b/advanced_alchemy/alembic/commands.py
@@ -244,5 +244,6 @@ class AlembicCommands:
             },
         )
         self.config = AlembicCommandConfig(**kwargs)
-        self.config.set_main_option("script_location", self.sqlalchemy_config.alembic_config.script_location)
+        if not self.config.get_main_option("script_location"):
+            self.config.set_main_option("script_location", self.sqlalchemy_config.alembic_config.script_location)
         return self.config


JacobCoffee (approved):

cofin:
@chaush-server thanks for the fix.  Can you explain a bit about how this problem manifests itself and get the tests passing?

chaush-server:
> @chaush-server thanks for the fix. Can you explain a bit about how this problem manifests itself and get the tests passing?

Hi, when I try to specify script_location in alembic.ini, when initializing alembic with litestar database init, it just creates the migrations folder, although script_location is specified in alembic.ini
I don't quite understand why the tests don't pass, so I'm sorry if I was stupid somewhere.

cofin:
> > @chaush-server thanks for the fix. Can you explain a bit about how this problem manifests itself and get the tests passing?
> 
> Hi, when I try to specify script_location in alembic.ini, when initializing alembic with litestar database init, it just creates the migrations folder, although script_location is specified in alembic.ini I don't quite understand why the tests don't pass, so I'm sorry if I was stupid somewhere.

You should not (only) set the `script_location` in the `alembic.ini` if you want to use the full Litestar integration,  you should set it in the `AlembicConfig` configuration of the plugin.  It's designed to pull your config from the SQLAlchemy plugin.

PR #247, opened 2024-08-21
feat: add `TZ` validator to `AuditModel`

### LINKED ISSUE #244: Enhancement: include support for sqladmin
### Summary

When using `DateTimeUTC` on a model, creating instances via the sqladmin UI fails b/c the datetime instance produced by the form and passed to sqlalchemy is naive.

See https://github.com/peterschutt/sqladmin-litestar-plugin/issues/22 and https://github.com/peterschutt/sqladmin-litestar-plugin/pull/30

Maybe those could be upstreamed into this lib as this problem won't be unique to using sqladmin with Litestar.

### Basic Example

```py
from __future__ import annotations

from advanced_alchemy.base import UUIDAuditBase
from sqladmin import ModelView
from sqlalchemy import Column, String

class Entity(UUIDAuditBase):
    my_column = Column(String(10))

class EntityAdmin(ModelView, model=Entity): ...
```
Would become
```py
from __future__ import annotations

from advanced_alchemy.base import UUIDAuditBase
from advanced_alchemy.somewhere import AuditModelView
from sqlalchemy import Column, String

class Entity(UUIDAuditBase):
    my_column = Column(String(10))

class EntityAdmin(AuditModelView, model=Entity): ...
```

### Drawbacks and Impact

More stuff to maintain and carry around in this library.

### Unresolved questions

_No response_

cofin:
Fixes #244

## Description

-

### DIFF
--- a/advanced_alchemy/base.py
+++ b/advanced_alchemy/base.py
@@ -18,6 +18,7 @@ from sqlalchemy.orm import (
     mapped_column,
     orm_insert_sentinel,
     registry,
+    validates,
 )
 from sqlalchemy.orm.decl_base import _TableArgsType as TableArgsType  # pyright: ignore[reportPrivateUsage]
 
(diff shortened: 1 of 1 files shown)


cofin:
@peterschutt I think the validator is a good idea.  Is this sufficient to cover your use case?

Alc-Alc (approved):

peterschutt:
> @peterschutt I think the validator is a good idea. Is this sufficient to cover your use case?

Yeah, the issue was that there was no way to force sqladmin to send through an aware timestamp, so having the base model assume naive == UTC covers it.

Cheers

PR #251, opened 2024-09-05
feat: add optional nanoid base models

winstxnhdw:
## Description

This PR adds an optional Nano ID base model integration.


winstxnhdw:
Hey @cofin, could you check if this CI failure is related?

cofin:
@winstxnhdw Thanks for this.  sorry i missed this over the weekend.  I'll take a look and review this as soon as i can.

cofin:
you'll probably want to run `rm pdm.lock && make install`.  It'll generate a new lock for that includes the nanoid deps.

JacobCoffee (approved):
Could we have some usage documentation around this maybe? In general, adding bits via PR here and there would greatly reduce 1.0 stable releases burden

winstxnhdw:
Definitely, but I am not sure where. There doesn’t seem to be any foundational structure for the docs yet.

PR #252, opened 2024-09-08
feat: support columns using sqlalchemy's `declared_attr` decorator

bfrascher:
## Description

Before this patch, using an `SQLAlchemyDTO` with a table having attributes using `declared_attr` would raise an `ImproperConfigurationError`.

For the uses of `declared_attr` and why it is useful, please see the sqlalchemy documentation and in particular the following two pages:

- https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.declared_attr
- https://docs.sqlalchemy.org/en/20/orm/declarative_mixins.html

Due to the added code path in the
`advanced_alchemy.extensions.litestar.dto.parse_type_from_element` function, ruff complained about "too many return statements".  This patch silences that lint for that function, so that it can focus on the added logic to support `declared_attr` and not restructure the function to work around the lint.


cofin (approved):
LGTM.  Thanks for the change

PR #255, opened 2024-09-24
feat: enable `MERGE` in `postgres` and `oracle`

cofin:
## Description

-


.gitignore
```diff
@@ -168,3 +168,6 @@ cython_debug/
 
 # generated changelog
 /docs/changelog.md
+.zed
+.cursorrules
+.cursorignore
```
JacobCoffee, line 173: just my latest thoughts but I feel like IDE specifics could go into a global gitignore (https://sebastiandedeyne.com/setting-up-a-global-gitignore-file) instead of per project to keep it lean.

No fuss now, though! :)

cofin:
I'm not exactly satisfied with how this is shaping up.  I'm going to park this PR for a bit.  We'll revisit this after we get through 1.0

PR #258, opened 2024-09-28
feat: remove unfinished merge operation

cofin:
## Description

It turns out that the existing merge code in the `operations` folder will require a bit more work before it's ready to be used in the repository.  

I've left the Merge PR open with this full code and I'm proposing we remove some of these long standing unused references.  We can re-add this when it's ready to be used again.


Alc-Alc (approved):
One could say, merge got *unmerged*

PR #268, opened 2024-10-21
fix(sanic): bind session to session class instead of to the session maker

### LINKED ISSUE #267: Bug: Sanic extension improperly binds session maker dependency
### Description

The dependency for session maker was defined, and then the dependency for session overwrites it with a session maker as the type.  This seems non-ideal -- you can't get the session maker and when you ask for the session maker you get a session object.

It feels like this was a mistake in the sanic extension definition.

_No response_

### MCVE

```python
session_config = AsyncSessionConfig(expire_on_commit=False)
sqlalchemy_config = SQLAlchemyAsyncConfig(
    connection_string=CONNECTION_STRING,
    session_config=session_config,
)  # Create 'db_session' dependency.
alchemy = SanicAdvancedAlchemy(sqlalchemy_config=sqlalchemy_config)

Extend.register(alchemy)
Extend(app)

@app.get()
async def health(self, request, session_maker: async_sessionmaker):
    print(session_maker)
    return json({"status": "ok"}, 200)
```

### Steps to reproduce

```bash
1. Start up server
2. Make request
3. Check `async_sessionmaker`
4. Observe `None`
```

### Screenshots

_No response_

### Logs

_No response_

### Package Version

v0.20.0

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

imnotjames:
## Description

binds session into sanic extension as expected

in the original code, session maker was defined and then the dependency for session overwrites it with a session maker as the type.  this seems non-ideal -- you can't get the session maker and when you ask for the session maker you get a session object

instead, this looks at the sessionmaker `class_` property for adding the sanic dependency

## Closes

fixes #267

### DIFF
--- a/advanced_alchemy/extensions/sanic.py
+++ b/advanced_alchemy/extensions/sanic.py
@@ -96,6 +96,10 @@ class SanicAdvancedAlchemy(Extension, Generic[EngineT, SessionT, SessionMakerT])
             else self.sqlalchemy_config.get_engine()
         )
         self.session_maker = self.sqlalchemy_config.create_session_maker()
+
+        session_maker = cast("SessionMakerT", self.session_maker)
+        self.session_class = session_maker.class_
+
         self.app: Sanic
 
     async def _do_commit(self, session: Session | AsyncSession) -> None:
@@ -181,7 +185,7 @@ class SanicAdvancedAlchemy(Extension, Generic[EngineT, SessionT, SessionMakerT])
                 self.get_sessionmaker_from_request,
             )
             bootstrap.add_dependency(
-                type(self.session_maker),
+                self.session_class,
                 self.get_session_from_request,
             )
 


advanced_alchemy/extensions/sanic.py
```diff
@@ -96,6 +97,14 @@ def __init__(
             else self.sqlalchemy_config.get_engine()
         )
         self.session_maker = self.sqlalchemy_config.create_session_maker()
+
+        if self.sqlalchemy_config.session_config.class_ is not Empty:
+            self.session_class = self.sqlalchemy_config.session_config.class_
+        elif isinstance(self.session_maker, async_sessionmaker):
+            self.session_class = AsyncSession
+        else:
+            self.session_class = Session
```
imnotjames, line 106: I don't think this is exactly right.. a session maker can create an async session.  However, I don't know a better way to determine this right now.

imnotjames: Ah, wait!  The session maker has the `class_` defined, I think..  In theory, that should ALWAYS have the correct value.  Bonus: We don't have to move imports around.

```suggestion
        else:
            self.session_class = self.session_maker.class_
```

imnotjames:
Should this also update the example code?  If we do this, the example code that binds session explicitly is no longer needed.

advanced_alchemy/extensions/sanic.py
imnotjames: Much better.

cofin (approved):

PR #273, opened 2024-10-28
fix: use `collection.abc.Sequence` instead of `typing.Sequence`

cofin:
## Description

-


Alc-Alc (approved):
I assume this is due to ruff?

PR #276, opened 2024-11-01
feat: improved typing for DTOData

cofin:
## Description

-

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ install: destroy clean								## Install the project, dependencies, and pre-comm
 
 .PHONY: clean
 clean: 												## Cleanup temporary build artifacts
-	@echo "=> Cleaning working directory"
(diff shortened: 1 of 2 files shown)


vikigenius:
Didn't notice this PR sorry about that. I created another one.

Even though the types are compatible now this still doesn't actually fix the issue because of ordering. In my PR https://github.com/litestar-org/advanced-alchemy/pull/277

I got rid of the fallback altogether since it is only ever used for type_checking and you gain nothing by having it if the package is not installed anyways.

cofin:
> Didn't notice this PR sorry about that. I created another one.
> 
> Even though the types are compatible now this still doesn't actually fix the issue because of ordering. In my PR #277
> 
> I got rid of the fallback altogether since it is only ever used for type_checking and you gain nothing by having it if the package is not installed anyways.

I'm going to close this in favor of yours.  Thanks for working on this.

Issue #265, opened 2024-10-09
Bug: Typing issues in ModelDictT


vikigenius:
### Description

pyright complains when I try to assign `DTOData[User]` to `ModelDictT[User]`

```
/home/void/Projects/Python/litestartest/src/litestartest/controller.py
  /home/void/Projects/Python/litestartest/src/litestartest/controller.py:35:45 - error: Argument of type "DTOData[User]" cannot be assigned to parameter "data" of type "ModelDictT[User]" in function "create"
    Type "DTOData[User]" is incompatible with type "ModelDictT[User]"
      "DTOData[User]" is incompatible with "Dict[str, Any]"
      "DTOData[User]" is incompatible with "User"
      "DTOData[User]" is incompatible with "Struct"
      "DTOData[User]" is incompatible with protocol "Struct"
        "__struct_fields__" is not present
      "DTOData[User]" is incompatible with "BaseModel"
      "DTOData[User]" is incompatible with protocol "BaseModel"
    ... (reportArgumentType)
1 error, 0 warnings, 0 informations
```
You can't really see the error because pyright is supressing the last lines. But if I remove BaseModel from one of the union types you cans see the actual error

```
      "litestar.dto.data_structures.DTOData" is incompatible with "advanced_alchemy.service.typing.DTOData" (reportArgumentType)
```

You usually can't reproduce this error until you install pydantic. Because without pydantic installed in strict mode

you get

```
  /home/void/Projects/Python/litestartest/src/litestartest/service.py:10:42 - error: Type of "ModelDictT" is partially unknown
    Type of "ModelDictT" is "type[Dict[str, Any]] | type[ModelT@ModelDictT] | type[msgspec.Struct] | type[advanced_alchemy.service.typing.Struct] | Unknown | type[BaseModel] | type[DTOData[ModelT@ModelDictT]]" (reportUnknownVariableType)
```

Which suppresses the other issue.

I think in general the way ModelDictT is typed with try except does not actually seem to work properly.

### URL to code causing the issue

https://github.com/vikigenius/litestartest/tree/main

### Steps to reproduce

Run pyright or basedpyright on repository
Verify that `reportUnknownVariableType` is being produced in strict mode (because pydantic is not installed).

Install pydantic

Run pyright or basedpyright again.

This should cause the

```
/home/void/Projects/Python/litestartest/src/litestartest/controller.py
[...]

PR #277, opened 2024-11-02
feat: Improved typing of `ModelDictT`

vikigenius:
## Description

Fixes typing issues in service

## Closes

https://github.com/litestar-org/advanced-alchemy/issues/265

This still doesn't solve the problem of UnknownVariableType if the subtypes of ModelDictT are not installed (eg: Pydantic)
But at least it solves the problem of incompatibilities when they are installed

### DIFF
--- a/advanced_alchemy/service/typing.py
+++ b/advanced_alchemy/service/typing.py
@@ -42,7 +42,7 @@ try:
     # this is from pydantic 2.8.  We should check for it before using it.
     from pydantic import FailFast  # pyright: ignore[reportAssignmentType]
 
-    PYDANTIC_USE_FAILFAST: Final[bool] = True
+    PYDANTIC_USE_FAILFAST: Final[bool] = False
 except ImportError:
     PYDANTIC_USE_FAILFAST: Final[bool] = False  # type: ignore # pyright: ignore[reportConstantRedefinition,reportGeneralTypeIssues]  # noqa: PGH003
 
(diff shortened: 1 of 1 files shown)


vikigenius:
@cofin  looks like the failfast checks fail because I changed

```python
try:
    # this is from pydantic 2.8.  We should check for it before using it.
    from pydantic import FailFast  # pyright: ignore[reportAssignmentType]

    PYDANTIC_USE_FAILFAST: Final[bool] = False
```

to

```python

try:
    # this is from pydantic 2.8.  We should check for it before using it.
    from pydantic import FailFast  # pyright: ignore[reportAssignmentType]

    PYDANTIC_USE_FAILFAST: Final[bool] = True
```

I am changing it back, but I guess I am confused about why it's always False and what purpose is it actually serving?

cofin (approved):

cofin:
> @cofin looks like the failfast checks fail because I changed
> 
> ```python
> try:
>     # this is from pydantic 2.8.  We should check for it before using it.
>     from pydantic import FailFast  # pyright: ignore[reportAssignmentType]
> 
>     PYDANTIC_USE_FAILFAST: Final[bool] = False
> ```
> 
> to
> 
> ```python
> try:
>     # this is from pydantic 2.8.  We should check for it before using it.
>     from pydantic import FailFast  # pyright: ignore[reportAssignmentType]
> 
>     PYDANTIC_USE_FAILFAST: Final[bool] = True
> ```
> 
> I am changing it back, but I guess I am confused about why it's always False and what purpose is it actually serving?

It's currently set to be permanently disabled.  It was enabled for a short time after this feature came out, but I would intermittently get exceptions when using it.  Because of this, I disabled it until I had time to get a bit more familiar with the feature (and for it to mature a bit).

PR #284, opened 2024-11-10
feat: lambda statement optimizations

cofin:
## Description

-


provinzkraut (approved):

advanced_alchemy/filters.py
```diff
@@ -202,14 +202,14 @@ def append_to_lambda_statement(
...
-                track_on=[id(values), id(model)],
+                track_on=[self.values, self.field_name, model],
             )
         values = self.values
         return statement.add_criteria(
             lambda s: s.where(field.in_(values)),
             track_bound_values=True,
             track_closure_variables=False,
-            track_on=[id(values), id(model)],
+            track_on=[self.values, self.field_name, model],
```
provinzkraut: If we track on `self.values`, I think this would cause a bug if any item in there isn't hashable 👀

advanced_alchemy/filters.py
```diff
@@ -252,14 +252,14 @@ def append_to_lambda_statement(
                 lambda s: s.where(any_(values) != field),  # type: ignore[arg-type]
                 track_bound_values=True,
                 track_closure_variables=False,
-                track_on=[id(values), id(model)],
+                track_on=[self.values, self.field_name, model],
```
provinzkraut: Same issue with hashing probably

advanced_alchemy/filters.py
```diff
@@ -252,14 +252,14 @@ def append_to_lambda_statement(
...
             )
         values = self.values
         return statement.add_criteria(
             lambda s: s.where(field.notin_(values)),
             track_bound_values=True,
             track_closure_variables=False,
-            track_on=[id(values), id(model)],
+            track_on=[self.values, self.field_name, model],
```
provinzkraut: Same issue with hashing probably

PR #303, opened 2024-12-01
feat: additional loader configurations and re-organization

cofin:
## Description

-

### DIFF
--- a/advanced_alchemy/mixins/audit.py
+++ b/advanced_alchemy/mixins/audit.py
@@ -1,0 +1,29 @@
+from __future__ import annotations
+
+from datetime import datetime, timezone
+
+from sqlalchemy.orm import Mapped, mapped_column, validates
+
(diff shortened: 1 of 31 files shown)


advanced_alchemy/mixins/audit.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+from datetime import datetime, timezone
+
+from sqlalchemy.orm import Mapped, declarative_mixin, mapped_column, validates
+
+from advanced_alchemy.types import DateTimeUTC
+
+
+@declarative_mixin
```
Alc-Alc, line 10: why was this added for all mixins? AFAIK this is not needed in 2.0

cofin: I added these to be consistent across all mixins.  We had one with them and the rest without.

We can remove them all of none of them are needed; but I'd just like it to be consistent.

PR #307, opened 2024-12-04
feat!: allow override of `registry` for declarative bases

cofin:
## Description

This PR allows you to resuse the existing DeclarativeBase subclasses with different metadata objects.

This enables better support for multi-database configurations with models targets for each specific database.

It:
- Creates a singleton registry object that tracks Metadata's for models
- uses the `__bind_key__` syntax commonly found in `flask-slqlalchemy` and other SQLAlchemy libraries

As part of this, there are a few (minor) breaking changes:
- The custom subclass for `AlembicCommands` has been removed.  It is no longer required and will make things easier for fastapi integration
- The Alembic configuration now longer has a `template_metadata` key


advanced_alchemy/base.py
```diff
@@ -291,167 +295,226 @@ def create_registry(
...
+    __abstract__ = True
+    __metadata_registry__: MetadataRegistry = MetadataRegistry()
+    __bind_key__: str | None = None
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        bind_key = getattr(cls, "__bind_key__", None)
+        if bind_key is not None:
+            if bind_key not in cls.__metadata_registry__:
+                cls.__metadata_registry__[bind_key] = MetaData(naming_convention=convention)
+            cls.metadata = cls.__metadata_registry__[bind_key]
```
Alc-Alc, line 357:
```suggestion
        if bind_key is not None:
            cls.metadata = cls.__metadata_registry__.get(bind_key)
```

Correct me if I am wrong, if we are going with the current implementation for `MetadataRegistry.get`, that already does this for you.

Alc-Alc (approved):

Issue #326, opened 2025-01-09
chore(ci): remove pinned version


cofin:
https://github.com/litestar-org/advanced-alchemy/blob/d59a52ecb7d276aaf105f856590f8282243cc49c/.python-version#L1

PR #329, opened 2025-01-10
chore: remove `.python-version`

cofin:
Removes the `.python-version` file that was erroneously added to the repository


cofin:
Closes #326

PR #333, opened 2025-01-11
docs: Add doc for creating a custom model

Kumzy:
Created a section to define a custom model. 

Maybe the example is not correct but could not think of a specific use case, if you have any I can update this.


cofin (approved):
LGTM

PR #335, opened 2025-01-11
docs: add types documentation

Kumzy:
Add documentation part for the types

It's missing more insight for the EncryptedString I think

### DIFF
--- a/docs/usage/types.rst
+++ b/docs/usage/types.rst
@@ -12,7 +12,7 @@ A timezone-aware DateTime type that ensures UTC timezone handling in the databas
 .. code-block:: python
 
     from advanced_alchemy.types import DateTimeUTC
-    
+
     class MyModel:
         created_at = Column(DateTimeUTC)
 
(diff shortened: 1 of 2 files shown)


docs/usage/types.rst
```diff
@@ -0,0 +1,187 @@
...
+    class User:
+        id = Column(GUID, primary_key=True)
+        created_at = Column(DateTimeUTC)
+        password = Column(EncryptedString(key="secret-key"))
+        preferences = Column(JsonB)
+
+Using Types with Alembic
+------------------------
+
+When using these types with Alembic migrations, you need to properly configure your ``script.py.mako`` template. The key is to make the custom types available through the ``sa`` namespace that Alembic uses.
```
cofin, line 142:
```suggestion
If you are not using Advanced Alchemy's built-in `alembic` templates, you need to properly configure your ``script.py.mako`` template. The key is to make the custom types available through the ``sa`` namespace that Alembic uses.
```

cofin (approved):
LGTM

PR #337, opened 2025-01-11
feat: add ``soft delete`` mixin

Kumzy:
Creating a mixin for soft delete.

Do you think we need more in the Mixin?

Missing tests + repository changes + documentation


hasansezertasan:
I love the idea, soft deletes rules!

That said, I'd love AA to have deeper support for it. Such as deletion-aware repository/service methods like get_one, list, etc.

PR #343, opened 2025-01-16
chore: don't clear pre-commit cache

sloria:
## Description

running `make install` blew away my _entire_ pre-commit cache (!) which i don't think was the intention.


cofin (approved):

cofin:
Thanks for this!  I did not intend to leave this in there.

PR #344, opened 2025-01-16
fix: disabled `timezone` in alembic.ini

Harshal6927:
Disabled `timezone` in alembic.ini to fix `alembic.util.exc.CommandError: Can't locate timezone: UTC` error while applying migrations

Reference:
https://alembic.sqlalchemy.org/en/latest/tutorial.html#editing-the-ini-file


cofin (approved):
LGTM

cofin:
Just to note, this is likely a Windows specific thing.

Harshal6927:
> Just to note, this is likely a Windows specific thing.

In that case, can someone from LiteStar server test it out?

cofin:
> > Just to note, this is likely a Windows specific thing.
> 
> In that case, can someone from LiteStar server test it out?

Yeah, I use this very config on Mac and Linux often.  There are also tests for the commands in the test suite (but I'd have to review how comprehensive each of the alembic tests are).

PR #345, opened 2025-01-16
docs: clarifications to services and modeling docs

sloria:
## Description

while reading through the example code in the Services usage docs, i had a hard time understanding what method calls were coming from where due to the lack of imports and a few incorrect usages. this fixes those, as well as adds a few more links

### DIFF
--- a/docs/usage/frameworks/fastapi.rst
+++ b/docs/usage/frameworks/fastapi.rst
@@ -2,7 +2,7 @@
 FastAPI Integration
 ===================
 
-Advanced Alchemy provides integration with FastAPI through its repository and service patterns.
+Advanced Alchemy's repository and service patterns work well within FastAPI applications.
 
 Basic Setup
 -----------
--- a/docs/usage/frameworks/litestar.rst
+++ b/docs/usage/frameworks/litestar.rst
@@ -380,12 +380,11 @@ Initializing a new project
 
 If you would like to initial set of alembic migrations, you can easily scaffold out new templates to setup a project.
 
-Assuming that you are using the default configuration for the SQLAlchemy configuration, you can run the following.
+Assuming that you are using the default configuration for the SQLAlchemy configuration, you can run the following to initialize the migrations directory.
 
-.. code-block:: bash
+.. code-block:: shell-session
 
-    # Initialize migrations directory
-    litestar database init ./migrations
+    $ litestar database init ./migrations
 
 If you use a different path than `./migrations`, be sure to also set this in your SQLAlchemy config.  For instance, if you'd like to use `./alembic`:
 
@@ -397,12 +396,11 @@ If you use a different path than `./migrations`, be sure to also set this in you
         ),
     )
 
-And then run the following:
+And then run the following to initialize the migrations directory:
 
-.. code-block:: bash
+.. code-block:: shell-session
 
-    # Initialize migrations directory
-    litestar database init ./alembic
+    $ litestar database init ./alembic
 
 You will now be configured to use the alternate directory for migrations.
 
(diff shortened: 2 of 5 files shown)


docs/usage/frameworks/litestar.rst
```diff
@@ -397,12 +396,11 @@ If you use a different path than `./migrations`, be sure to also set this in you
...
 
-And then run the following:
+And then run the following to initialize the migrations directory:
 
-.. code-block:: bash
+.. code-block:: shell-session
 
-    # Initialize migrations directory
-    litestar database init ./alembic
+    $ litestar database init ./alembic
```
sloria, line 403: note: without the `$` pygments doesn't highlight correctly, and the text becomes too low contrast

docs/usage/frameworks/litestar.rst
```diff
@@ -397,12 +396,11 @@ If you use a different path than `./migrations`, be sure to also set this in you
         ),
     )
 
-And then run the following:
+And then run the following to initialize the migrations directory:
 
-.. code-block:: bash
+.. code-block:: shell-session
 
-    # Initialize migrations directory
```
sloria, line 404: i removed the comments to so that the copy button only copies the shell commands

docs/usage/frameworks/fastapi.rst
```diff
@@ -2,7 +2,7 @@
 FastAPI Integration
 ===================
 
-Advanced Alchemy provides integration with FastAPI through its repository and service patterns.
+Advanced Alchemy's repository and service patterns work well within FastAPI applications.
```
sloria, line 5: iiuc, there's no integration code within advanced-alchemy for FastAPI. the docs are just showing that repositories and services can work within FastAPI apps, correct?

docs/usage/services.rst
```diff
@@ -89,37 +89,39 @@ Services provide high-level methods for common operations:
             data=data,
             auto_commit=True,
         )
-        return post_service.to_schema(post)
+        return post_service.to_schema(post, schema_type=PostResponse)
 
 Complex Operations
 -------------------
 
-Services can handle complex business logic involving multiple repositories:
```
sloria, line 97: afaict, the example code doesn't actually use multiple repositories. i think it'd be good to eventually have an example of multiple repos, but that can happen later

docs/usage/services.rst
```diff
@@ -66,17 +67,16 @@ Services provide high-level methods for common operations:
 
 .. code-block:: python
 
-    async def create_post_with_tags(
+    async def create_post(
         post_service: PostService,
         data: PostCreate,
     ) -> PostResponse:
         """Create a post with associated tags."""
-        # Service automatically validates input using PostCreate schema
```
sloria, line 74: the service doesn't actually validate input, right? the data passed in to the service methods are assumed to be valid?

docs/usage/frameworks/fastapi.rst
cofin: @sloria That's correct. As soon as I figure out what's blocking these flask CLI tests (they are only failing in the CI right now), i'll move over to the FastAPI cli.

docs/usage/services.rst
cofin: Correct, this is a good update.

docs/usage/services.rst
cofin: This is also correct.  I guess I forgot to come back to this one.

cofin (approved):
@sloria Thanks for taking the time on this.  It's greatly appreciated!

docs/usage/frameworks/fastapi.rst
sloria: no urgency on my end for FastAPI cli support (not using it myself), though i'm keen to to try out litestar :) really great work on litestar and advanced-alchemy. advanced-alchemy would have be super useful on my last work project ❤️

PR #352, opened 2025-01-18
feat(fastapi): refactored integration with CLI support

cofin:
Refactored the Starlette and FastAPI integration to support multiple configurations and sessions.  Additionally, FastAPI will now have the database commands automatically registered with the FastAPI CLI.


advanced_alchemy/extensions/starlette/config.py
```diff
@@ -0,0 +1,344 @@
...
+
+    Returns:
+        str: A unique key name.
+    """
+    i = 0
+    while True:
+        if not hasattr(app.state, key):
+            return key
+        key = f"{key}_{i}"
+        i += i
```
Alc-Alc, line 52: this is not `i += 1`?

tests/unit/test_extensions/test_fastapi.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+@pytest.fixture()
+def mock_rollback(mocker: MockerFixture, config: AnyConfig) -> MagicMock:
+    if isinstance(config, SQLAlchemySyncConfig):
+        return mocker.patch("sqlalchemy.orm.Session.rollback")
+    return mocker.patch("sqlalchemy.ext.asyncio.AsyncSession.rollback")
+
+
+@pytest.mark.xfail(
```
Alc-Alc, line 78: the reason says its skipping the version, so why not use `pytest.mark.skipif`?

Alc-Alc, line 78: if you are using skip, you can probably move the fixture to a module level scope and remove decorating individual tests
```
pytestmark = pytest.mark.skipif(
    condition=sys.version_info < (3, 8),
    reason="Certain versions of Starlette and FastAPI are stated to still support 3.8, but there are documented incompatibilities on various versions that have not been yanked.  Skipping on 3.8 for now until this is resolved.",
)
```

advanced_alchemy/extensions/fastapi/__init__.py
```diff
@@ -0,0 +1,35 @@
+"""FastAPI extension for Advanced Alchemy.
+
+This module provides FastAPI integration for Advanced Alchemy, including session management,
+database migrations, and service utilities.
+
+"""
+
+from advanced_alchemy import base, exceptions, filters, mixins, operations, repository, service, types, utils
```
Alc-Alc, line 8: why are we re exporting these in fastapi? I don't think there is much convenience in doing `from advanced_alchemy.fastapi import *` over the normal `from advanced_alchemy import *`

cofin (commented):
> why are we re exporting these in fastapi? I don't think there is much convenience in doing `from advanced_alchemy.fastapi import *` over the normal `from advanced_alchemy import *` 

It's for convenience so you don't need to import from several places with AA.  It's done like this for Litestar and Flask as well.

cofin (commented):
> if you are using skip, you can probably move the fixture to a module level scope and remove decorating individual tests
> ```
> pytestmark = pytest.mark.skipif(
>     condition=sys.version_info < (3, 8),
>     reason="Certain versions of Starlette and FastAPI are stated to still support 3.8, but there are documented incompatibilities on various versions that have not been yanked.  Skipping on 3.8 for now until this is resolved.",
> )
> ``` 

It's probably important for me to point out that these are all of the existing test cases for starlette, not any that i've written yet.

provinzkraut (approved):

PR #358, opened 2025-01-21
feat: add hash to config classes

### LINKED ISSUE #357: Enhancement: Allow config classes to be hashable
### Summary

Currently, `SQLAlchemySyncConfig` and `SQAlchemyAsyncConfig` classes are _not_ hashable (maybe due to SQLAlchemy attributes). 

I would like to be able to "register" these classes into a `svcs` Registry, but I cant do that currently because `svcs` requires a class to be hashable:

From svcs docs:
>But the types must be hashable because they’re used as keys in a lookup dictionary.

I'd like to introduce a `__hash__` function for both of those classes, so that they will be compatible with `svcs`.

### Basic Example

If I have a config item that looks like this:

```python
config = SQLAlchemySyncConfig(
    engine_instance=sync_engine, session_maker=sync_session_factory, alembic_config=sync_alembic_config
)
```

I want to be able to register `config` with services. I would do that like this:
```python
registry = svcs.Registry()

registry.register_value(SQLAlchemySyncConfig, config)

container = svcs.Container(registry)
```

This, in turn, allows me to retrieve the class when needed during specific calls:
```python
config = container.get(SQLAlchemySyncConfig)
```

It's hard to see the value in this small example, but using `svcs` as the central manager of my services (both for setup and teardown, which is handled automatically), it is quite useful.

### Drawbacks and Impact

In order to force a dataclass to be hashable, one _could_ include `unsafe_hash=True`, that could have some unintended consequences.

However, adding a unique hash function might be trivial, and I'm not aware of any immediate impact/drawback for current use. 

One could add a `__hash__` function that includes `uuid3` or `uuid5`, along with one or more attributes from the actual class.

### Unresolved questions

_No response_

tataraba:
## Description

Adds hash function to `SQLAlchemySyncConfig` and `SQLAlchemyAsyncConfig` classes.

## Closes
Closes #357

### DIFF
--- a/advanced_alchemy/config/asyncio.py
+++ b/advanced_alchemy/config/asyncio.py
@@ -71,6 +71,12 @@ class SQLAlchemyAsyncConfig(GenericSQLAlchemyConfig[AsyncEngine, AsyncSession, a
     The configuration options are documented in the Alembic documentation.
     """
 
+    def __hash__(self) -> int:
+        return super().__hash__()
+
+    def __eq__(self, other: object) -> bool:
+        return super().__eq__(other)
+
     @asynccontextmanager
     async def get_session(
         self,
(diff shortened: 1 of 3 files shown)


cofin (approved):
LGTM

Issue #367, opened 2025-01-25
Bug: startup and shutdown events don't run in fastapi apps with lifespan


PokkaKiyo:
### Description

In FastAPI (and Starlette), the startup and shutdown events will not run if the lifespan event is specified. The lifespan method is the recommended way to perform startup and shutdown tasks in FastAPI (and Starlette).

The advanced alchemy strategy to hook into these apps seems to rely on adding "startup" and "shutdown" event handlers.
However, if the user defines a custom lifespan in the FastAPI constructor, the startup and shutdown event handlers will not run.
This would primarily mean that tables won't be created in the startup (if create_all was specified), and the database engine won't be closed in the shutdown. 

#### References

- https://fastapi.tiangolo.com/advanced/events/#alternative-events-deprecated
- https://github.com/encode/starlette/blob/0.45.3/starlette/routing.py#L598-L610

### MCVE

```python

import asyncio
import contextlib
from contextlib import asynccontextmanager

from advanced_alchemy.base import BigIntBase
from advanced_alchemy.extensions.fastapi import AdvancedAlchemy, SQLAlchemyAsyncConfig
from fastapi import FastAPI


class Author(BigIntBase): ...


async def fn(alchemy: AdvancedAlchemy):
    async with alchemy.with_async_session() as session:
        session.add(Author())
        await session.commit()


app1 = FastAPI()  # app1 does not have lifespan set
alchemy1 = AdvancedAlchemy(
    config=SQLAlchemyAsyncConfig(
        connection_string="sqlite+aiosqlite:///test1.sqlite",
        create_all=True,
    ),
    app=app1,
)


@asynccontextmanager
async def lifespan(app: FastAPI):
    try:
        yield
    finally:
        pass


app2 = FastAPI(lifespan=lifespan)  # app2 has lifespan set
alchemy2 = AdvancedAlchemy(
    config=SQLAlchemyAsyncConfig(
        connection_string="sqlite+aiosqlite:///test2.sqlite",
        create_all=True,
    ),
    app=app2,
)


async def test_fn():
    from uvicorn import Config, Server

    server1 = Server(Config(app="__main__:app1", port=8001))
    task1 = asyncio.create_task(server1.serve())
    await asyncio.sleep(2)
    server2 = Server(Config(app="__main__:app2", port=8002))
    task2 = asyncio.create_task(server2.serve())
    await asyncio.sleep(2)

    try:
        await fn(alchemy1)
    except Exception as e:
[...]

PR #368, opened 2025-01-25
fix: use `lifespan` context manager in Starlette and FastAPI

cofin:
Modifes the Starlette and FastAPI integrations to use the `lifespan` context manager instead of the `startup`\`shutdown` hooks.  If the application already has a lifespan set, it is wrapped so that both execute.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
       - id: unasyncd
         additional_dependencies: ["ruff"]
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.9.2"
+    rev: "v0.9.3"
     hooks:
       # Run the linter.
     - id: ruff
@@ -32,7 +32,7 @@ repos:
     - id: ruff-format
       types_or: [ python, pyi ]
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.3.0
+    rev: v2.4.0
     hooks:
       - id: codespell
         exclude: "uv.lock|examples/us_state_lookup.json"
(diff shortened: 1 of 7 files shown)


cofin:
Closes #367

PokkaKiyo:
@cofin If you don't mind me adding, the changes you made to AdvancedAlchemy seem like it would run each config's on_startup/on_shutdown twice if the user does not define a custom lifespan, but only once if the user does define lifespan.

That's probably fine for create_all and engine.dispose, but maybe the add_event_handler functions in the config classes should be removed if this isn't intended?

cofin:
> @cofin If you don't mind me adding, the changes you made to AdvancedAlchemy seem like it would run each config's on_startup/on_shutdown twice if the user does not define a custom lifespan, but only once if the user does define lifespan.
> 
> That's probably fine for create_all and engine.dispose, but maybe the add_event_handler functions in the config classes should be removed if this isn't intended?
> 
> Also the outer shutdown doesn't run when hit with an exception 😅

Hah.  Is it obvious I don't use FastAPI daily? :)

cofin:
> @cofin If you don't mind me adding, the changes you made to AdvancedAlchemy seem like it would run each config's on_startup/on_shutdown twice if the user does not define a custom lifespan, but only once if the user does define lifespan.
> 
> That's probably fine for create_all and engine.dispose, but maybe the add_event_handler functions in the config classes should be removed if this isn't intended?
> 
> Also the outer shutdown doesn't run when hit with an exception 😅

Your suggestions were spot on.  I think I've *finally* addressed all of the issues on this one.  Thanks again.

provinzkraut (approved):

PokkaKiyo:
Thank you!

PR #369, opened 2025-01-26
docs: cli

Kumzy:
Add documentation for the cli part

### DIFF
--- a/docs/usage/cli.rst
+++ b/docs/usage/cli.rst
@@ -61,6 +61,16 @@ The CLI can be invoked using Python's module syntax:
     # or with uv
     uv run -m advanced_alchemy --help
 
+Global Options
+--------------
+
+The following options are available for all commands:
+
+--config TEXT        Required. Dotted path to SQLAlchemy config(s)
+--bind-key TEXT     Optional. Specify which SQLAlchemy config to use
+--no-prompt         Optional. Skip confirmation prompts
+--verbose          Optional. Enable verbose output
+
 Available Commands
 ------------------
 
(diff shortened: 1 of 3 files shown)


docs/usage/cli.rst
```diff
@@ -0,0 +1,275 @@
...
+
+        .. code-block:: bash
+            :caption: Using `PDM <https://pdm.fming.dev/>`_
+
+            pdm add advanced-alchemy[cli]
+
+    .. tab-item:: Poetry
+
+        .. code-block:: bash
+            :caption: Using `Poetry <https://python-poetry.org/>`_
```
Alc-Alc, line 48: not sure why others are even here, but ok

docs/usage/cli.rst
```diff
@@ -0,0 +1,275 @@
...
+   * - ``--no-prompt``
+     - Optional. Skip confirmation prompts
+   * - ``--verbose``
+     - Optional. Enable verbose output
+
+
+Config
+------
+
+Here is an example of what **config** would looks like.
```
Alc-Alc, line 86:
```suggestion
Here is an example of what **config** looks like.
```

Alc-Alc (commented):
Nice. As an aside, CSS is broken though, not sure if it's only me

docs/usage/cli.rst
Kumzy: I kept the same thing as on the get started (installation) page

Alc-Alc (approved):
can revisit the phrasing / wording later if needed

Issue #365, opened 2025-01-24
Enhancement: Way to hook into the database CLI


DataDaoDe:
### Summary

I would like to extend the database command to be able to do something like the following:

```python
class DatabaseCLIPlugin(CLIPluginProtocol):
    def on_cli_init(self, cli: Group) -> None:
        db_group = get_alchemy_group()

        @db_group.command(name="create-database", help="Uses the config connection_string to create a new database if none exists")
        def create_database():
            ....
```

I can't seem to figure out a way to hook into the cli and extend the advanced alchemy command though. I can roll my db cli command and integrate that with litestar, but I'd rather not have two commands i.e. one called `database` and then my custom one called `db`.

Does anyone have any ideas?

Alternatively or additionally, is there any interest in taking up two new commands: `create-db` and `drop-db` which create and drop the database respectively. I tend to have these in a lot of my projects.

Thanks! 

### Basic Example

```python

class CLIPlugin(CLIPluginProtocol):
    def on_cli_init(self, cli: Group) -> None:
        console = get_console()

        # for create/drop commands we cannot be connected to the db we want to create/drop
        app_db_url = urlparse(settings.DATABASE_URL)
        app_db_name = app_db_url.path.replace("/", "")
        pg_db_url_result = app_db_url._replace(path="/postgres")
        pg_db_url = urlunparse(pg_db_url_result)

        async_engine = create_async_engine(
            url=pg_db_url,
            execution_options={"isolation_level": "AUTOCOMMIT"},
        )

        @cli.command(name="db-create")
        def create_database(app: Litestar) -> None:
            async def _create_db():
                async with async_engine.begin() as conn:
                    try:
                        sql = "CREATE DATABASE {} ENCODING '{}'".format(
                            app_db_name, "utf-8"
                        )
                        await conn.execute(text(sql))
                    except DatabaseError as err:
                        console.print(err)

            asyncio.run(_create_db())

        @cli.command(name="db-drop")
        def drop_database(app: Litestar) -> None:
            async def _drop_db():
                async with async_engine.connect() as conn:
                    try:
                        conn = await conn.execution_options(
                            isolation_level="AUTOCOMMIT"
                        )

                        # We must first disconnect all users from the database we are dropping.
                        version = conn.dialect.server_version_info
                        pid_column = "pid" if (version >= (9, 2)) else "procpid"
                        drop_user_conn_sql = """SELECT pg_terminate_backend(pg_stat_activity.{pid_column})
[...]

PR #371, opened 2025-01-27
feat: add database create and drop commands

DataDaoDe:
## Description

This code adds two generic database commands, `create` and `drop` for creating an new database and dropping an existing one respectively. It was briefly discussed some here: #365 . A lot of the code was taken from [SQLAlechemy Utils database.py](https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/functions/database.py) 

Currently, it still needs integration testing, so far I've only tested it locally with sqlite.

### DIFF
--- a/advanced_alchemy/cli.py
+++ b/advanced_alchemy/cli.py
@@ -402,8 +402,17 @@ def add_migration_commands(database_group: Group | None = None) -> Group:  # noq
     @database_group.command(name="create", help="Create a new database.")
     @bind_key_option
     @no_prompt_option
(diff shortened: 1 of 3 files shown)


cofin:
@DataDaoDe you may hit a wall on some of these engines (maybe Cockroach and Spanner, specifically).  Let's just highlight these places so we c an see how many of them there are.

cofin:
@DataDaoDe I'm going to close this PR, but feel free to re-open it and link to the v2 milestone if there's interest in finishing it up for all drivers.

PR #397, opened 2025-02-22
feat: add Enum to default type decoders

cofin:
Extends the default `msgspec` type decoders to handle Enum types by converting them to their underlying value during serialization


Kumzy (approved):
LGTM

PR #401, opened 2025-02-26
feat(litestar): Use `SerializationPlugin`

provinzkraut:
Use `SerializationPlugin` instead of `SerializationPluginProtocol`


Kumzy (approved):
LGTM

PR #421, opened 2025-03-18
fix(repo): logic correction for window function

cofin:
Corrects the logic for using a count with a window function.


provinzkraut (approved):
Should we have a test for this case?

cofin:
> Should we have a test for this case?

We actually had one but it was checking for both conditions.  I've corrected it now.

PR #426, opened 2025-04-01
fix: remove accidental litestar import

cofin:
Remove an incorrect import of `console` from `litestar.cli._utils` and replace it with a correct import from `rich`. This change ensures proper functionality without unnecessary dependencies.

### DIFF
--- a/advanced_alchemy/alembic/utils.py
+++ b/advanced_alchemy/alembic/utils.py
@@ -20,7 +20,7 @@ async def drop_all(engine: "Union[AsyncEngine, Engine]", version_table_name: str
         version_table_name: The name of the version table.
         metadata: The metadata object containing the tables to drop.
     """
-    from rich import get_console  # noqa: PLC0415
+    from rich import get_console
 
     console = get_console()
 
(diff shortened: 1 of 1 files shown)


Kumzy (commented):
Hello,
The ruff linting rule failed. 

Also, is there some case where we want the user to be able to use methods inside utils like ``drop_all`` (without going through the AA cli command) ? If so we should:
-  manage if the cli optional dependencies is installed
- raise a warning ?
- make the console output happening only if the cli dependency is installed
- ignore all the above?

cofin:
> Hello, The ruff linting rule failed.
> 
> Also, is there some case where we want the user to be able to use methods inside utils like `drop_all` (without going through the AA cli command) ? If so we should:
> 
> * manage if the cli optional dependencies is installed
> * raise a warning ?
> * make the console output happening only if the cli dependency is installed
> * ignore all the above?

Good call.  I added a quick check to raise the correct MissingDependencyError.

Kumzy (approved):
LGTM

PR #429, opened 2025-04-05
feat(litestar): server side session backend

cofin:
Implements a server side session backend using SQLAlchemy.  Works with an Async or Sync configuration.


advanced_alchemy/utils/time.py
```diff
@@ -0,0 +1,8 @@
+import datetime
+import sys
+
+
+def get_utc_now() -> datetime.datetime:  # pragma: no cover
+    if sys.version_info >= (3, 11):
+        return datetime.datetime.now(datetime.UTC)
+    return datetime.datetime.utcnow().replace(tzinfo=None)  # pyright: ignore  # noqa: DTZ003
```
provinzkraut, line 8: You can always do `datetime.datetime.now(tz=datetime.timezone.UTC)`, it's backwards compatible

cofin: I don't think `datetime.timezone.UTC` exists in 3.9

provinzkraut: You can use `datetime.timezone.utc` instead

PR #442, opened 2025-04-14
fix: make `SentinelMixin` compatible with `MappedAsDataclass`

winstxnhdw:
`MappedAsDataclass` is a mixin introduced in SQLAlchemy 2.0. It introduces massive DX improvements to SQLAlchemy by introducing dataclass type validation to SQLAlchemy models. However, this mixin is incompatible with SQLAlchemy's recommended method of implementing a sentinel column as written in their [documentation](https://docs.sqlalchemy.org/en/20/core/connections.html#configuring-sentinel-columns).

This PR fixes this incompatibility as suggested by the SQLAlchemy maintainer in this [discussion](https://github.com/sqlalchemy/sqlalchemy/discussions/12519#discussioncomment-12804658).


winstxnhdw:
Hmm.. `init=False` is not allowed if the class is not using `MappedAsDataclass`. That's really annoying.

cofin (approved):

PR #445, opened 2025-04-14
chore(docs): update session docs

Harshal6927:
## Description

- Change `alchemy.provide_session` to `sqlalchemy_config.provide_session`
- Use `ASGIMiddleware`
- Add example for `get_session`

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -75,6 +75,17 @@ html.dark .title-with-logo .brand-text {
   text-shadow: var(--brand-text-glow);
 }
 
+/* Button container wrapping */
+.buttons.wrap {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 0.5rem;
+}
+
+.buttons.wrap .btn-no-wrap {
+  flex: 0 0 auto;
+}
+
 /* Large screens */
 @media (min-width: 1200px) {
   html[class] .title-with-logo .brand-text {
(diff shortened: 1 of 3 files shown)


docs/usage/frameworks/litestar.rst
```diff
@@ -303,20 +303,21 @@ You can access the database session from the controller by using the `db_session
         plugins=[alchemy],
     )
 
-Sessions in Middleware
-^^^^^^^^^^^^^^^^^^^^^^
```
cofin, line 307: The idea of this section was to show how to access session that was already attached into the session's scope.

This is just showing how to create a custom middleware.  Is there something else you were trying to highlight?

docs/usage/frameworks/litestar.rst
cofin, line 1: Instead of demonstrating this with a middleware.  Can we show this as an guard?  i think the idea was to highlight the usage of `provide_session` so that it gets the session that was already in the request scope.

Harshal6927: Sure, should I show `provide_session` example with guard and keep the `get_session` as it is?

docs/_static/custom.css
```diff
@@ -75,6 +75,17 @@ html.dark .title-with-logo .brand-text {
   text-shadow: var(--brand-text-glow);
 }
 
+/* Button container wrapping */
```
cofin, line 78: Can we move this into a separate PR that specifies what it's fixing?

Harshal6927: Sure

PR #448, opened 2025-04-19
chore: update session docs

Harshal6927:
## Description

- Change `alchemy.provide_session` to `sqlalchemy_config.provide_session`
- Add example for `get_session`


docs/usage/frameworks/litestar.rst
```diff
@@ -327,20 +330,81 @@ Dependency injection is not available in middleware. Instead, you can create a n
...
+                    ],
+                ),
+            ],
+        ),
+    )
+
+
+    @get("/")
+    async def hello() -> str:
+        return "Hello, world!"
```
cofin, line 402: I don't think we should reference other plugins in our examples

docs/usage/frameworks/litestar.rst
```diff
@@ -303,15 +303,18 @@ You can access the database session from the controller by using the `db_session
         plugins=[alchemy],
     )
 
-Sessions in Middleware
-^^^^^^^^^^^^^^^^^^^^^^
+Sessions in Application
+^^^^^^^^^^^^^^^^^^^^^^^
+
+You can use either ``provide_session`` or ``get_session`` to get session instances in your application. This is useful for components that need to access the session outside of a controller.
```
cofin, line 309:
```suggestion
You can use either ``provide_session`` or ``get_session`` to get session instances in your application. Each of these functions are useful for providing sessions in various places within your application, whether you are in the request/response scope or not.
```

docs/usage/frameworks/litestar.rst
```diff
@@ -303,15 +303,18 @@ You can access the database session from the controller by using the `db_session
...
 
-Dependency injection is not available in middleware. Instead, you can create a new session using the `provide_session` method:
+- ``provide_session`` provides a session instance from request state if it exists, or creates a new session if it doesn't.
```
cofin, line 311:
```suggestion
- ``provide_session`` provides a session instance from request state if it exists, or creates a new session if it doesn't, while `get_session` always returns a new instance from the session maker.

`provide_session` is useful in places where you are already in the request/response context such as guards and middleware.
`get_session` is useful anywhere outside of the request lifecycle in your application.  This includes command line tasks and background jobs.

```

docs/usage/frameworks/litestar.rst
cofin: Take a look at swapping this out with a CLI plugin example.  

Great work so far though!

cofin:
I added a bit more details about when/why you would use each one.  Feel free to reword my additions as needed though.

Harshal6927:
I tried switching to CLI example for `get_session` but it started giving coroutine error because `get_session` is `async` generator and `cli.command()` only uses `sync` callable.

cofin:
Take a look at the edits I've made here.  This should get you on the right track re: the CLI plugin.

PR #452, opened 2025-04-21
feat: PasswordHash field type

cofin:
Implements a PasswordHash field type with multiple supported backends.

Includes built-in backends for:
- `passlib`
- `argon2`
- `pwdlib`


advanced_alchemy/types/password_hash/argon2.py
```diff
@@ -0,0 +1,97 @@
...
+
+if TYPE_CHECKING:
+    from sqlalchemy import BinaryExpression, ColumnElement
+
+PasswordHasher = None  # type: ignore[var-annotated]
+InvalidHash = None  # type: ignore[var-annotated]
+VerifyMismatchError = None  # type: ignore[var-annotated]
+with contextlib.suppress(ImportError):
+    from argon2 import PasswordHasher  # type: ignore[assignment]
+    from argon2.exceptions import InvalidHash, VerifyMismatchError  # type: ignore[assignment]
```
provinzkraut, line 17: If this module only contains the argon2 related code, no need to suppress these import errors here; You shouldn't import this code if you don't have the lib installed

PR #465, opened 2025-04-30
fix(types): correct type hints for with_for_update to ForUpdateParameter

PrinterOMG:
## Description

This change fixes the type hint for the `with_for_update` parameter in the repositories. Currently, the type hint is `Optional[bool]`, but according to the method's docstring, `with_for_update` is described as: "_indicating FOR UPDATE should be used, or may be a dictionary containing flags to indicate a more specific set of FOR UPDATE flags for the SELECT_". This implies that a dictionary may be passed, but the current type hint doesn’t reflect that.

If we trace where `with_for_update` is passed, it ends up unchanged in session.refresh in SQLAlchemy, which has the type hint `ForUpdateParameter = Union["ForUpdateArg", None, bool, Dict[str, Any]]`, and that’s what I updated the type hints in our code to match.

However, it’s possible that this isn’t what was originally intended, and I’m not sure if this is the correct fix. There are other ways to address the issue — which approach would be preferable?


cofin (approved):
Good catch.  This looks correct to me.  Thanks for the PR!

PR #474, opened 2025-05-07
docs: fix docs and type hint of repository.add_many

PokkaKiyo:
## Description

- Updates the imports in the litestar example to working ones (AA 1.4, litestar 2.16)
- Repository example uses create_many, but that method doesn't exist (add_many exists)
- ~~The return type of add_many is always a list since the input argument is a list~~
- Service.add_many doesn't have uniquify

### DIFF
--- a/docs/usage/repositories.rst
+++ b/docs/usage/repositories.rst
@@ -139,7 +139,7 @@ Create Many
         repository = PostRepository(session=db_session)
 
         # Create posts
-        return await repository.create_many(
+        return await repository.add_many(
             [Post(title=title, content=content, author_id=author_id) for title, content, author_id in data],
             auto_commit=True
         )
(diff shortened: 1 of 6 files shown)


PokkaKiyo:
Went through the docs over the last few days and spotted a couple of outdated parts, hopefully fixed them!

I'm not too sure about my change to the return type of add_many looking at the other types... would be happy to revert them.

cofin:
Thanks for the changes.  Re: the import update from `litestar.plugins.sqlalchemy` to `advanced_alchemy`. I recognize that these probably had type errors due to the way the re-exported submodules are handled in Python.  I've got a fairly large fix that should be getting merged into Litestar to fix this.

Once this is working correctly, I'll probably end up changing these references back, but I think it's better to have something working now.

cofin (approved):

docs/usage/repositories.rst
```diff
@@ -130,16 +130,16 @@ Bulk Operations
 
 Repositories support efficient bulk operations:
 
-Create Many
+Add Many
 -----------
```
cofin, line 134:
```suggestion
Add Many
--------
```

PR #498, opened 2025-07-12
fix: correct typing issue in `litestar` example

yuuuxt:
## Description

- fix according to https://github.com/litestar-org/litestar-fullstack/blob/0996536189cf48c0b7ce8c2c2600cb439968a377/src/app/domain/accounts/controllers/users.py#L49


cofin (approved):
LGTM

PR #579, opened 2025-10-15
fix: ensure `has_dict_attribute` checks for `__dict__` attribute

haggqvist:
## Description

The previous implementation used `isinstance` against the `DictProtocol` type returned `True` for any object.

For example:

```python
from advanced_alchemy.service.typing import has_dict_attribute
has_dict_attribute(1)         # True
has_dict_attribute([1, 2, 3]) # True
```

The `isinstance` call is replaced with `hasattr` (which should also be faster). With this change, I believe the `DictProtocol` class could also be removed, but I kept the changes to a minimum.


cofin (approved):
Thanks, @haggqvist.  LGTM

PR #580, opened 2025-10-15
fix: surface FileObject session errors; align commit/rollback semantics

barapa:
- Previously save/delete failures were only logged, so callers believed commits succeeded when storage ops had already failed.
- Sync commit: processes saves sequentially, then deletes. Logs failures with tracebacks and re-raises. Ignores FileNotFoundError on delete. Stops on first save error. Clears internal state only on full success.
- Async commit: runs saves and deletes concurrently via asyncio.gather. Logs each failure with exc_info. Raises the first real Exception, and lets BaseException (e.g., CancelledError) bubble. Attempts deletes even if a save fails. Clears state only on full success.
- Rollback (sync/async): deletes only files saved during the current transaction. Ignores FileNotFoundError. Logs and re-raises other errors. Clears state after processing.
- Tracking: records successful saves in _saved_in_transaction for targeted rollback cleanup; state is retained on error to allow inspection/retry.

Tests
- Sync/async save and delete error paths, including FileNotFoundError handling and BaseException bubbling.
- Override semantics: save cancels prior delete; delete cancels prior save.
- State behavior: clear on success; retain on error.
- Async specifics: delete still attempted even if a save fails; raises first failing Exception in input order; logs include exc_info for stack traces.

### DIFF
--- a/advanced_alchemy/types/file_object/session_tracker.py
+++ b/advanced_alchemy/types/file_object/session_tracker.py
@@ -27,7 +27,7 @@ class FileObjectSessionTracker:
 
         Args:
             raise_on_error: If True, raise exceptions on file operation failures.
-                           If False, log warnings and continue (legacy behavior).
+                            If False, log warnings and continue.
 
         Internal structures:
         - ``pending_saves``: ``FileObject -> data`` to be saved on commit


barapa:
Paraphrasing @Harshal6927:
With this change, errors are now raised on failed uploads. So this should be considered a breaking change.

advanced_alchemy/config/common.py
```diff
@@ -181,6 +182,20 @@ class GenericSQLAlchemyConfig(Generic[EngineT, SessionT, SessionMakerT]):
     This is a listener that will automatically save and delete :class:`FileObject <advanced_alchemy.types.file_object.FileObject>` instances when they are saved or deleted.
 
     Disable if you plan to bring your own save/delete mechanism for these columns"""
+    file_object_raise_on_error: bool = False
```
provinzkraut: IMO this could also default to `True`. While that would be breaking in some sense, I'd say the previous behaviour is a bug, so fixing that bug with how it should actually behave would be fine

cofin (approved):

Harshal6927 (approved):

advanced_alchemy/types/file_object/session_tracker.py
```diff
@@ -17,8 +22,20 @@
...
     """Tracks FileObject changes within a single session transaction."""
 
-    def __init__(self) -> None:
-        """Initialize the tracker."""
+    def __init__(self, raise_on_error: bool = False) -> None:
+        """Initialize empty tracking state.
+
+        Args:
+            raise_on_error: If True, raise exceptions on file operation failures.
+                           If False, log warnings and continue (legacy behavior).
```
cofin:
```suggestion
                            If False, log warnings and continue.
```

advanced_alchemy/config/common.py
```diff
@@ -181,6 +181,12 @@ class GenericSQLAlchemyConfig(Generic[EngineT, SessionT, SessionMakerT]):
     This is a listener that will automatically save and delete :class:`FileObject <advanced_alchemy.types.file_object.FileObject>` instances when they are saved or deleted.
 
     Disable if you plan to bring your own save/delete mechanism for these columns"""
+    file_object_raise_on_error: bool = True
```
provinzkraut: Just curious, is there *ever* a situation where you want to not raise in this case?

PR #585, opened 2025-10-18
feat: add support for SQLAlchemy func expressions in filter classes

### LINKED ISSUE #519: Enhancement: add support for `sqlalchemy.func` type
### Summary

Add support for `sqlalchemy.func` type so that, for example, using `func.random()` does not raise a linting error.

### Basic Example

Current:
```py
OrderBy(func.random()) # Argument of type "random" cannot be assigned to parameter "field_name" of type "str" in function "__init__" "random" is not assignable to "str"
```

Image

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

cofin:
## Summary

Adds support for SQLAlchemy func() expressions in filter classes to eliminate type checker errors when using database functions like `func.random()` or `func.lower()`.

Closes #519

## The Problem

Type checkers currently raise errors when using SQLAlchemy func expressions in filters:

```python
OrderBy(func.random())  # Type error: expected 'str', got 'random'
```

The `field_name` parameter was typed as `str`, but `func.random()` returns a `FunctionElement` (subclass of `ColumnElement`).

## The Solution

Updated type signatures to accept `Union[str, ColumnElement[Any], InstrumentedAttribute[Any]]` in filter classes. The implementation already worked - this just fixes the type annotations.

## Key Features

- Works with all SQLAlchemy func expressions (`func.random()`, `func.lower()`, etc.)
- 100% backward compatible - existing string-based code unchanged
- Type safe - both mypy and pyright pass in strict mode
- Supports all 8+ database backends (each has different function names)

## Example

```python
from sqlalchemy import func
from advanced_alchemy.filters import OrderBy

# Random ordering
users = await repository.list(OrderBy(func.random()))

# Case-insensitive sorting
users = await repository.list(OrderBy(func.lower(User.name)))

# Backward compatible - still works
users = await repository.list(OrderBy("created_at", sort_order="desc"))
```

## Changes

- Updated `OrderBy`, `BeforeAfter`, `OnBeforeAfter`, `CollectionFilter`, `NotInCollectionFilter`, `ComparisonFilter`
- Enhanced `_get_instrumented_attr()` to handle new types
- Mock repositories extract field names from `InstrumentedAttribute`, raise helpful error for func expressions (can't execute SQL in-memory)
- Added integration tests for func expressions

Note: Different databases use different function names (PostgreSQL: `func.random()`, MySQL: `func.rand()`, SQL Server: `func.newid()`).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
       - id: unasyncd
         additional_dependencies: ["ruff"]
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.14.1"
+    rev: "v0.14.2"
     hooks:
       # Run the linter.
       - id: ruff
--- a/uv.lock
+++ b/uv.lock
@@ -36,7 +36,8 @@ dependencies = [
 
 [package.optional-dependencies]
 argon2 = [
-    { name = "argon2-cffi" },
+    { name = "argon2-cffi", version = "23.1.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.10'" },
+    { name = "argon2-cffi", version = "25.1.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.10'" },
 ]
 cli = [
     { name = "rich-click" },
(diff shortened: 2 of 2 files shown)


advanced_alchemy/repository/memory/_sync.py
```diff
@@ -210,36 +210,60 @@ def _exclude_unused_kwargs(self, kwargs: dict[str, Any]) -> dict[str, Any]:
...
+        Raises:
+            RepositoryError: If a ColumnElement (func expression) is used with mock repository
+        """
+        if isinstance(field, str):
+            return field
+        if isinstance(field, InstrumentedAttribute):
+            # Extract the attribute name from InstrumentedAttribute
+            return field.key
+        # ColumnElement (func expressions) cannot be executed in-memory
+        msg = "func expressions and column elements are not supported in mock repositories (in-memory filtering)"
```
provinzkraut: Are we sure these are `func` expressions at this point? It could also be any other unsupported type iiuc

provinzkraut: Yes, I think that would be good. Maybe report the actual type that was received and is unsupported. e.g. `{type(field)} columns are not supported...`

PR #608, opened 2025-11-12
chore: modify readme for correct dark/light mode of image and fix the…

Kumzy:
This PR fixes:
- Correct image displaying for light and dark mode following [GitHub recommendation](https://github.blog/developer-skills/github/how-to-make-your-images-in-markdown-on-github-adjust-for-dark-mode-and-light-mode/) 
- Fix the table with badges by removing the unused column


Harshal6927:
We'll also have to update the [pypi_readme](https://github.com/litestar-org/advanced-alchemy/blob/main/tools/pypi_readme.py)

Kumzy:
> We'll also have to update the [pypi_readme](https://github.com/litestar-org/advanced-alchemy/blob/main/tools/pypi_readme.py)

It should not impact the pypi_readme.py as the changes are mostly between the markers and for the table, its the removing of one column

PR #629, opened 2025-12-11
chore(docs): fix project docs link in README.md

apejcic:
## Description

-


Harshal6927:
Thank you for the PR. We are currently addressing some domain management issues, and the documentation is still served from the old domain. We plan to move everything to the new https://advanced-alchemy.dev as soon as possible, so this change is not required at this time.

PR #664, opened 2026-01-24
fix(litestar DTO): Ensure ORM descriptor fields are not evaluated during DTO creation

### LINKED ISSUE #646: Bug: Problem with certain hybrid properties starting with 1.8.0
### Description

I've hit a problem with out codebase when upgrading advanced_alchemy from 1.7.0
I have the following hybrid property which no longer works with 1.8.0
```
    @hybrid_property
    def children_presets(self) -> list['Preset']:
        return [child for child in self.all_children_presets if not child.hidden]
```
I have created a minimum example that works fine with 1.7.0 but cannot start in 1.8.0

_No response_

### MCVE

```python
from litestar import Litestar, get
from sqlalchemy.ext.asyncio import AsyncSession
from sqlalchemy.orm import Mapped
from sqlalchemy import select
from advanced_alchemy.extensions.litestar import (
    AsyncSessionConfig,
    SQLAlchemyAsyncConfig,
    SQLAlchemyPlugin,
    SQLAlchemyDTO,
    base,
)
from sqlalchemy.ext.hybrid import hybrid_property
import uvicorn

class Entity(base.UUIDBase):
    name: Mapped[str]
    @hybrid_property
    def name_letters(self) -> list[str]:
        return list(set([a for a in self.name]))

EntityDTO = SQLAlchemyDTO[Entity]

@get("/entities", return_dto=EntityDTO)
async def get_all_entities(db_session: AsyncSession) -> list[Entity]:
    result = await db_session.execute(select(Entity))
    entities = result.scalars().all()
    return list(entities)


session_config = AsyncSessionConfig(expire_on_commit=False)
alchemy_config = SQLAlchemyAsyncConfig(
    connection_string="sqlite+aiosqlite:///:memory:",
    session_config=session_config,
    create_all=True,
)

async def create_sample_entities() -> None:
    async with alchemy_config.get_session() as db_session:
        entities = [
            Entity(name="Entity One"),
            Entity(name="Entity Two"),
        ]
        
        db_session.add_all(entities)
        await db_session.commit()

app = Litestar(
    route_handlers=[get_all_entities],
    plugins=[SQLAlchemyPlugin(config=alchemy_config)],
    on_startup=[create_sample_entities],
)

if __name__ == "__main__":
    uvicorn.run(app)
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

"In the format of: `SCREENSHOT_DESCRIPTION`"

### Logs

```bash

```

### Package Version

1.8.0

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

provinzkraut:
Fix #646.

When processing property fields on SQLAlchemy models for the Litestar DTO, there was sometimes unexpected behaviour caused by the inspecting code evaluating the ORM descriptors such as `hybrid_property`. 

Fix this behaviour by skipping all known ORM fields (as reported by the ORM), and not using `inspect.getmembers`.

### DIFF
--- a/advanced_alchemy/extensions/litestar/dto.py
+++ b/advanced_alchemy/extensions/litestar/dto.py
@@ -1,5 +1,4 @@
 # ruff: noqa: C901
-import inspect as stdlib_inspect
 import logging
 from collections.abc import Collection, Generator
 from collections.abc import Set as AbstractSet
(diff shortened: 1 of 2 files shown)


cofin (approved):
LGTM - much more elegant than what I tried to do.

PR #678, opened 2026-02-22
feat(release): add pre-release version support

cofin:
## Summary

- Add PEP 440 pre-release (alpha/beta/rc) support to `bump-my-version` config
- Add `make pre-release version=X.Y.ZaN` Makefile target
- Update `tools/prepare_release.py` to validate pre-release versions and set the GitHub `prerelease` flag on draft releases
- Document pre-release workflow in `CONTRIBUTING.rst` and fix version examples in `docs/releases.rst`

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -145,3 +145,57 @@ Creating a new release
 10. **Publish to PyPI**
 
     - Approve the ``Latest Release`` workflow under ``Actions`` to publish the package to PyPI.
+
+Creating a pre-release
+----------------------
+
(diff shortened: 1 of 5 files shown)


cofin:
> have you thought about uv's? https://docs.astral.sh/uv/guides/package/#updating-your-version

We could probably use that for this project.  I actually hadn't looked at it.  I ported what we use in `litestar-vite`.  And I don't think `uv`'s method will work there.  The `bump-to-version` let's us change the version in multiple files based on a template.

I'll look into the UV version thing a bit more though.  It would be nice to simplify things.

Issue #684, opened 2026-02-27
Bug: `update()` missing `was_attribute_set()` guard on relationship loop which silently clears relationships during partial updates


sohbit:
### Description

Hey folks,

We recently upgraded `advanced_alchemy` to the latest (from `1.4.5` to `1.9.0b1` (commit `fee3dea`)) and noticed that there was a breaking change on how partial updates are being handled in regards to relationships.

When you do something like `repo.update(MyModel(id=existing_id, name="new name"))`, you'd expect only `name` to change. And for columns, that works perfectly. The `was_attribute_set()` guard added in #563 handles it. But the relationship loop right below it doesn't have the same guard, so SQLAlchemy's auto-initialized defaults (`None` for scalars, `[]` for collections) get treated as intentional values and written through to the DB.

**The result:** updating a single field can silently detach related records. In our case, updating a address' name was clearing its phone number association. Took a while to figure out because there's no error. It just quietly nulls things out.

Depending on whether the FK is nullable you get one of two outcomes:
- **Nullable FK**: relationship silently cleared: data loss with no error
- **Non-nullable FK**: `IntegrityError` crash

### Where the bug is

The column loop in `update()` does the right thing:

```python
for column in mapper.mapper.columns:
    # ...
    if not was_attribute_set(data, mapper, field_name):  # ✅ checks before writing
        continue
    # ...
```

But the relationship loop right below it skips that check:

```python
for relationship in mapper.mapper.relationships:
    if relationship.viewonly or relationship.lazy in {
        "write_only", "dynamic", "raise", "raise_on_sql",
    }:
        continue
    # ← no was_attribute_set() check ❌
    if (new_value := getattr(data, relationship.key, MISSING)) is not MISSING:
        # ...writes None or [] onto the existing instance
```

The problem is that `getattr(data, relationship.key, MISSING)` doesn't return `MISSING` for unset relationships. SQLAlchemy's descriptor auto-initializes them, so you get `None` or `[]` instead. The `is not MISSING` check passes, and the value gets written through.

### MCVE

```python
Tested and confirmed on `1.9.0b1` (commit `fee3dea`), SQLAlchemy 2.0.44, Python 3.14.


"""
Two independent scenarios:
  1. Nullable FK  → relationship silently cleared (data loss, no error)
  2. Non-nullable FK → IntegrityError crash
"""

import asyncio
from uuid import uuid4
from advanced_alchemy.base import UUIDAuditBase
[...]

PR #685, opened 2026-02-27
fix: add `was_attribute_set()` guard to relationship loop in `update()`

sohbit:
## Description

Partial updates using model instances (e.g. `Model(id=..., name='new')`) were silently clearing relationships because SQLAlchemy auto-initializes unset relationship attributes to `None/[]`, which then passed the 'is not MISSING' check and got written through to the DB.

This adds the same `was_attribute_set()` guard that already protects the column loop to the relationship loop, so only explicitly set relationships are copied during `update()`.

## Closes 

- #684

### DIFF
--- a/advanced_alchemy/repository/_async.py
+++ b/advanced_alchemy/repository/_async.py
@@ -2276,6 +2276,12 @@ class SQLAlchemyAsyncRepository(SQLAlchemyAsyncRepositoryProtocol[ModelT], Filte
                             # Skip relationships with incompatible lazy loading strategies
                             continue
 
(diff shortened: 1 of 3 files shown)


sohbit:
Hi @cofin,

let me know if I can do anything in order to fix the CI here. Seems to be unrelated to the changes from the PR. Tests were passing locally.

cofin:
@sohbit it may be a transient error.  I've triggered a retry and will keep an eye on it.

cofin (approved):
LGTM

PR #718, opened 2026-04-26
fix: docs preview workflow

Harshal6927:
## Description

- Fixes docs preview workflow

*Reference: https://github.com/litestar-org/litestar/blob/main/.github/workflows/docs-preview.yml*

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -10,9 +10,9 @@ jobs:
     if: ${{ github.event.workflow_run.conclusion == 'success' &&  github.event.workflow_run.event == 'pull_request' }}
     runs-on: ubuntu-latest
     permissions:
-      issues: write
+      actions: read
+      contents: read
       pull-requests: write
-      contents: write
 
     steps:
       - name: Check out repository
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-preview.yml
```diff
@@ -25,48 +28,66 @@ jobs:
...
+            exit 1
+          fi
+          echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
+
+      - name: Generate GitHub App Token
+        id: generate-token
+        uses: actions/create-github-app-token@v3
+        with:
+          client-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
```
Harshal6927: Need to set these secrets. @Kumzy can you let me know how can I generate these credentials?

cofin (approved):

.github/workflows/docs-preview.yml
Harshal6927: fyi I renamed `APP_ID` to `CLIENT_ID` and `APP_PRIVATE_KEY` to `CLIENT_PRIVATE_KEY`. So now we need to set `CLIENT_ID` and `CLIENT_PRIVATE_KEY`

Kumzy: Created the GH app and the secrets

Kumzy: Should we roll out the same changes to sqlspec and other projects that are using docs preview? @cofin

Kumzy (approved):

PR #732, opened 2026-05-12
test: migrate object storage tests to rustfs

cofin:
Replaces MinIO-backed object storage test setup with pytest-databases RustFS, including explicit test bucket creation for S3 round trips.

Updates the local FileObject examples to use RustFS defaults instead of MinIO credentials.

<hr>
📚 Documentation preview: <a href="https://litestar-org.github.io/advanced-alchemy-docs-preview/732">https://litestar-org.github.io/advanced-alchemy-docs-preview/732</a>


examples/fastapi/fastapi_fileobject.py
```diff
@@ -37,12 +37,15 @@
 app = FastAPI()
 alchemy = AdvancedAlchemy(config=alchemy_config, app=app)
 document_router = APIRouter()
+# For local development, run RustFS on port 9000 and create the static-files bucket first.
```
provinzkraut: Should probably just say "an S3 compatible storage" instead pf specifically mention RustFS? Or at least just use RustFS as an example. Otherwise it might sound like RustFS *specifically* is required for this

examples/litestar/litestar_fileobject.py
```diff
@@ -24,12 +24,15 @@
 from advanced_alchemy.types.file_object.data_type import StoredObject
 
 # Object storage backend
+# For local development, run RustFS on port 9000 and create the static-files bucket first.
```
provinzkraut: Same comment

provinzkraut (approved):

PR #744, opened 2026-05-27
feat: add get_or_none as an alias for get_one_or_none

whyujjwal:
This PR adds a new method get_or_none() as an alias for the existing get_one_or_none() method in the repository class. It also adds a test case for the new alias.

### DIFF
--- a/advanced_alchemy/repository/_async.py
+++ b/advanced_alchemy/repository/_async.py
@@ -251,6 +251,40 @@ class SQLAlchemyAsyncRepositoryProtocol(FilterableRepositoryProtocol[ModelT], Pr
         **kwargs: Any,
     ) -> Optional[ModelT]: ...
 
+    async def get_or_none(
+        self,
+        *filters: Union[StatementFilter, ColumnElement[bool]],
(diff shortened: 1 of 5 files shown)


cofin:
@whyujjwal thanks for the PR.  Can you elaborate a bit more on why this needed?  Is this a personal preference or was there an issue you've attempted to correct?

whyujjwal:
Hi @cofin, thanks for the question.

This is primarily a developer experience improvement. While  is functionally correct,  is a more common and arguably more intuitive method name for this operation in many other libraries and ORMs. 

By providing it as an alias, the intention is to:
- Make the API easier to use for developers familiar with the  convention.
- Reduce cognitive load by providing a more guessable method name.

There wasn't a specific bug this was correcting, but rather it's a small enhancement to make the library more user-friendly.

cofin:
Thanks @whyujjwal, for the change.  However, at this time, I don't think we are going to go down this route.

If you feel there's a strong case for this, hop in our discord for a quick chat about it before you make additional changes.

PR #749, opened 2026-05-30
docs(litestar): correct and test server-side session backend docs

### LINKED ISSUE #745: Bug: Litestar session example does not work as expected
### Description

When trying out the [Litestar session example](https://advanced-alchemy.litestar.dev/latest/usage/frameworks/litestar.html#complete-example), and using the exact code as posted in the docs, I see behaviour that I didn't expect and can't explain:

 - The database `sessions.db` gets created, as expected
 - It contains a table `web_sessions`, as expected.
 - However, even though session data seems to be saved *somewhere*, the table is always empty.

_No response_

### MCVE

```python
from litestar import Litestar, get, post
from litestar.connection import ASGIConnection
from litestar.middleware.session.server_side import ServerSideSessionConfig
from advanced_alchemy.extensions.litestar import (
    AsyncSessionConfig,
    SQLAlchemyAsyncConfig,
    SQLAlchemyPlugin,
)

from advanced_alchemy.extensions.litestar.session import (
    SessionModelMixin,
    SQLAlchemyAsyncSessionBackend,
)

# Session model
class WebSession(SessionModelMixin):
    __tablename__ = "web_sessions"

# Database configuration
alchemy_config = SQLAlchemyAsyncConfig(
    connection_string="sqlite+aiosqlite:///sessions.db",
    session_config=AsyncSessionConfig(expire_on_commit=False),
    create_all=True,
)

# Session configuration
session_config = ServerSideSessionConfig(
    max_age=3600,  # 1 hour
)

# Session backend
session_backend = SQLAlchemyAsyncSessionBackend(
    config=session_config,
    alchemy_config=alchemy_config,
    model=WebSession,
)

# Routes
@get("/")
async def home(request: ASGIConnection) -> dict:
    username = request.session.get("username")
    return {"message": f"Hello {username}!" if username else "Hello stranger!"}

@post("/login")
async def login(request: ASGIConnection) -> dict:
    form = await request.form()
    username = form.get("username")

    if username:
        request.set_session({"username": username, "login_time": "now"})
        return {"message": f"Welcome {username}!"}

    return {"error": "Username required"}

@post("/logout")
async def logout(request: ASGIConnection) -> dict:
    request.clear_session()
    return {"message": "Logged out successfully"}

# Application
app = Litestar(
    route_handlers=[home, login, logout],
    plugins=[SQLAlchemyPlugin(config=alchemy_config)],
    middleware=[session_config.middleware],
)
```

### Steps to reproduce

```bash
 1. Use the code snippet from the docs
 2. Run the server
 3. Run `curl -d "username=test" http://127.0.0.1:8000/login`, observe that the server responds with a positive message
 4. Open the database, e.g. using `uvx --from sqlit-tui sqlit sqlite://sessions.db` and query the table `web_sessions`, observe that it is empty when it should not be.
```

### Screenshots

_No response_

### Logs

```bash

```

### Package Version

v1.10.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

cofin:
## Problem

The "Server-Side Session Backend" guide built a `SQLAlchemyAsyncSessionBackend` and then wired the app with `middleware=[session_config.middleware]`. But `ServerSideSessionConfig.middleware` hardcodes Litestar's built-in `ServerSideSessionBackend` (resolving storage via `app.stores.get(config.store)`, default `"sessions"`), so the constructed backend was **never used** — session data silently went to an in-memory store and the DB table stayed empty. (Verified against litestar 2.23.)

Because `usage/frameworks/litestar.rst` was in `NON_EXECUTABLE_DOCS`, the examples were never run, so this and several related inaccuracies shipped.

Fixes #745
Fixes #707

## Changes

**Docs — two correct integration styles**
- **Store-based (recommended):** register `SQLAlchemyStore` under the `"sessions"` store name and keep `session_config.middleware`. (This is what `litestar-fullstack-inertia` already does.)
- **Backend-based:** wire with `partial(SessionMiddleware, backend=session_backend)` instead of `session_config.middleware`.
- Added a reference page for the previously-undocumented `SQLAlchemyStore` / `StoreModelMixin`.

**Docs — further factual fixes surfaced once the examples actually ran**
- `SessionModelMixin` has no `created_at`/`updated_at` (it extends `UUIDv7Base`, not an audit base) — corrected the field list and the `CREATE TABLE` schema (added `sa_orm_sentinel`).
- `__table_args__` override used `super().__table_args__()` (the directive already returns a tuple) and referenced the nonexistent `created_at` column — corrected.
- `ServerSideSessionConfig` has no `https_only` field — corrected to `secure`.

**Tests / regression guards**
- Behavioral integration tests for both wiring patterns asserting session rows actually persist to the DB.
- Made the session-integration blocks **Sybil-executable** (added `SkipParser`, moved the doc to `EXECUTABLE_DOCS`, wrapped the remaining context-sharing blocks in skip regions) so the examples can't silently rot again.

<hr>
📚 Documentation preview: <a href="https://litestar-org.github.io/advanced-alchemy-docs-preview/749">https://litestar-org.github.io/advanced-alchemy-docs-preview/749</a>

### DIFF
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -4,7 +4,7 @@ import pytest
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, create_async_engine
 from sqlalchemy.orm import sessionmaker
 from sybil import Sybil
-from sybil.parsers.rest import PythonCodeBlockParser
+from sybil.parsers.rest import PythonCodeBlockParser, SkipParser
 
 EXECUTABLE_DOCS = (
     "usage/modeling/basics.rst",
@@ -16,6 +16,9 @@ EXECUTABLE_DOCS = (
     "usage/repositories/filtering.rst",
     "usage/database_seeding.rst",
     "usage/services.rst",
+    # litestar.rst is executable only for the session-integration blocks; all
+    # other code blocks are wrapped in ``.. skip: start``/``.. skip: end``.
+    "usage/frameworks/litestar.rst",
 )
 
 NON_EXECUTABLE_DOCS = (
@@ -23,7 +26,6 @@ NON_EXECUTABLE_DOCS = (
     "usage/cli.rst",
     "usage/frameworks/fastapi.rst",
     "usage/frameworks/flask.rst",
-    "usage/frameworks/litestar.rst",
     "usage/frameworks/sanic.rst",
     "usage/frameworks/starlette.rst",
     "usage/routing.rst",
@@ -47,7 +49,7 @@ async def db_session_fixture(engine: AsyncEngine) -> AsyncGenerator[AsyncSession
 
 
 pytest_collect_file = Sybil(
-    parsers=[PythonCodeBlockParser()],
+    parsers=[PythonCodeBlockParser(), SkipParser()],
     patterns=EXECUTABLE_DOCS,
     fixtures=["db_session", "engine"],
 ).pytest()
--- a/docs/reference/extensions/litestar/index.rst
+++ b/docs/reference/extensions/litestar/index.rst
@@ -56,3 +56,4 @@ Additional API References
     plugins
     cli
     session
+    store
--- a/docs/reference/extensions/litestar/store.rst
+++ b/docs/reference/extensions/litestar/store.rst
@@ -1,0 +1,24 @@
+=====
+Store
+=====
+
+.. currentmodule:: advanced_alchemy.extensions.litestar.store
+
+This module provides a SQLAlchemy-backed implementation of Litestar's
+:class:`Store <litestar.stores.base.Store>` protocol. Register a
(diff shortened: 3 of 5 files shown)


nils-werner:
Hm, the [complete example](https://advanced-alchemy.litestar.dev/latest/usage/frameworks/litestar.html#complete-example) implements the backend based solution, and only mentions the store based solution in the following paragraph. Yet on the other hand, the store based solution is mentioned as the recommended one? That seems a bit odd and confusing to me...

And does the store based solution support `delete_expired()` like the backend based one does? Judging from the source code it doesn't?

Also, at least when using the backend based solution, I noticed that the session data must still be used from memory somewhere, even though it seems to be saved to the database:

 1. I can `DELETE FROM` the entire table, and my users remain logged in
 2. If I reboot the server, all my users are logged out, even though the session data remains in the database.

PR #761, opened 2026-06-09
chore(docs): update intersphinx domain for msgspec

shenanigansd:
ref: https://github.com/msgspec/msgspec/issues/1035
prior art: https://github.com/litestar-org/litestar/pull/4838


cofin:
Thanks, @shenanigansd closing this as a duplicate.  The fix should be in main now though.

PR #774, opened 2026-07-21
feat(alembic): auto-enable PostgreSQL enum migration detection

### LINKED ISSUE #492: Enhancement: Built-in PostgreSQL Enum Migration Support using alembic-postgresql-enum
### Summary

First, I want to express my sincere appreciation for Litestar and all the incredible work the maintainers and contributors have put into this project. The framework has been a joy to work with.

### Problem Description

When using PostgreSQL as a database with SQLAlchemy models that contain enum columns, the migration system doesn't automatically detect changes to Python enum definitions.

**Current Behavior:**
- Initial migration generation works perfectly with `litestar database make-migrations`
- Adding new enum values to existing Python enums is not detected by subsequent migration runs
- Migrations are silently skipped, leaving the database schema out of sync

**Example Scenario:**

Initial model definition:
```python
class UserType(StrEnum):
    ADMIN = "admin"
    PRINCIPAL = "principal"
    DEPUTY = "deputy"
    DIRECTOR = "director"
    TEACHER = "teacher"

class User(base.BigIntAuditBase):
    __tablename__ = "user"
    
    external_id: Mapped[str] = mapped_column(unique=True, index=True)
    user_type: Mapped[UserType] = mapped_column(
        nullable=False, index=True, default=UserType.TEACHER
    )
```

After adding a new enum value:
```python
class UserType(StrEnum):
    ADMIN = "admin"
    PRINCIPAL = "principal"
    DEPUTY = "deputy"
    DIRECTOR = "director"
    TEACHER = "teacher"
    STUDENT = "student"  # New enum value
```

Running `litestar database make-migrations` after this change produces no migration file.

### Current Workaround

The issue can be resolved by installing the `alembic-postgresql-enum` package and adding `import alembic_postgresql_enum` to the Alembic `env.py` file. This enables proper PostgreSQL enum change detection.

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

hasansezertasan:
## Summary

Closes #492.

Alembic's autogenerate cannot detect changes to PostgreSQL native `ENUM` types on its own (for example, adding a value to an existing `StrEnum`). Those changes are silently skipped, leaving the schema out of sync. The established fix is the [`alembic-postgresql-enum`](https://pypi.org/project/alembic-postgresql-enum/) package, which registers the missing autogenerate comparators purely as an import side effect.

Since Advanced Alchemy ships its own migration `env.py` templates, this PR wires that support in directly so it works out of the box.

## Changes

- **`alembic/templates/{sync,asyncio}/env.py`** — Guarded `import alembic_postgresql_enum` in a `try`/`except ImportError`. It activates only when the package is installed and is a no-op otherwise, so no existing users break.
- **`pyproject.toml`** — New public `postgresql` extra (`pip install "advanced-alchemy[postgresql]"`) and the package added to the dev `postgres` group so CI exercises it.
- **`docs/usage/cli.rst`** — Documents the behavior under `make-migrations`, including how existing projects can enable it by adding the import to an already-generated `env.py`.
- **`tests/integration/test_alembic_commands.py`** — Asserts the generated `env.py` includes the conditional import.

## Notes

The generated template intentionally keeps an explicit `try`/`except` (rather than `contextlib.suppress`) since users read and edit these files; `SIM105` is ignored for the template path accordingly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<hr>
📚 Documentation preview: <a href="https://litestar-org.github.io/advanced-alchemy-docs-preview/774">https://litestar-org.github.io/advanced-alchemy-docs-preview/774</a>


cofin:
I'll take a review of this tonight.  I think we have to be careful to ensure this is not a breaking change for the 1.0 branch.  This would include the migrations scaffolding as well, which I think is the challenge.

PR #781, opened 2026-08-13
fix(filters): escape LIKE wildcards in SearchFilter values

ruicleite96:
`SearchFilter` and `NotInSearchFilter` gain an opt-in `escape_wildcards` flag that matches `%` and `_` in the search value literally instead of treating them as SQL wildcards. Default behaviour is unchanged.

- **Literal matching is opt-in.** With `escape_wildcards=False` (the default) the emitted SQL, bound parameters and cache keys are identical to before, so existing searches keep their results.
- **Escaping uses `/`, not a backslash.** SQLAlchemy renders a backslash into the `ESCAPE` clause doubled, which DuckDB rejects. `/` compiles cleanly on every dialect that supports `ESCAPE`.
- **Every entry point honours the flag.** The in-memory mock repository matches literally too, and the Litestar and FastAPI filter providers accept a `search_escape_wildcards` config key so query-string search can opt in.
- **Reusable helpers are exported.** `LIKE_ESCAPE_CHAR` and `escape_like_value()` are public alongside the filters.
- **Spanner has no `ESCAPE` clause**, so the flag must stay off there. SQL Server bracket patterns (`[abc]`) are unchanged either way.

## How you use it

```python
SearchFilter(field_name="title", value="50% off", escape_wildcards=True)   # literal "50% off"
SearchFilter(field_name="title", value="50% off")                          # "%" stays a wildcard

create_filter_dependencies({"search": "title", "search_escape_wildcards": True})
```

<hr>
📚 Documentation preview: <a href="https://litestar-org.github.io/advanced-alchemy-docs-preview/781">https://litestar-org.github.io/advanced-alchemy-docs-preview/781</a>


cofin:
@ruicleite96 i've made a few tweaks as well to this one.  The change should be non-breaking, so I've disabled the new behavior by default along with a few other tweaks.  

Thanks again for the changes.

PR #782, opened 2026-08-13
docs(filters): clarify EXISTS subquery correlation

ruicleite96:
## Problem

`ExistsFilter` describes a correlated subquery, but `Select.correlate()` only permits correlation; it cannot infer the join. The existing example omitted the condition connecting the subquery to the outer table, so it could match every outer row whenever any user matched.

## Change

- Document that `values` must include the correlation condition.
- Correct the example to join `User.organization_id` to `Organization.id`.
- Clarify that the OR-only example is not correlated by itself.

This is a documentation-only update; runtime behavior is unchanged.

## Tests

- `uv run pytest tests/integration/test_filters.py -k exists -m sqlite -q` (4 passed)


cofin:
@ruicleite96   Can we start with a simple documentation update instead of a warning?

ruicleite96:
@cofin Done — I removed the runtime warning and its tests, leaving only the documentation update that explains the required correlation condition and corrects the examples. I also updated the PR title and description to reflect the reduced scope.

PR #783, opened 2026-08-13
feat(filters): add nulls placement to OrderBy

ruicleite96:
`OrderBy` gains an optional `nulls` field, `"first"` or `"last"`, that pins where NULL rows sort. Left unset, the emitted SQL is exactly what it was before, so existing queries are unaffected.

- **Databases disagree by default.** PostgreSQL and Oracle sort NULLs last ascending and first descending; MySQL and SQLite do the opposite. The same `OrderBy` therefore pages differently per backend, and on PostgreSQL a descending sort over a nullable column fills the first page with empty values.
- **Native clause where the syntax exists.** PostgreSQL, SQLite, Oracle and the rest get `NULLS FIRST` / `NULLS LAST`, which an index on the column can still satisfy.
- **Emulated where it does not.** MySQL, MariaDB and SQL Server reject the clause. They sort NULL as the lowest value, so `asc` + `first` and `desc` + `last` compile to the plain ordering term. The other two placements get a leading `CASE WHEN col IS NULL` sort key.
- **Plays well with the ORM.** The ordering term is a `UnaryExpression`, so eager loading with `limit`/`offset`, aliased models, labels and the compiled-statement cache all behave as with a native `nulls_last()`.
- **Every entry point honours it.** The in-memory mock repository applies the placement, and the Litestar and FastAPI filter providers accept a `sort_nulls` config key.
- **SQL Server caveat.** Under `SELECT DISTINCT`, SQL Server requires every ORDER BY term in the select list, so only the two native placements work with `DISTINCT` there.

## How you use it

```python
OrderBy(field_name="published_at", sort_order="desc", nulls="last")

create_filter_dependencies({"sort_field": "published_at", "sort_nulls": "last"})
```

Related to #782, which addresses the `ExistsFilter` correlation issue independently.

<hr>
📚 Documentation preview: <a href="https://litestar-org.github.io/advanced-alchemy-docs-preview/783">https://litestar-org.github.io/advanced-alchemy-docs-preview/783</a>


cofin (approved):
LGTM - thanks for the change

cofin:
Actually, there are some edge cases that need to be addressed here.  I'm going to just push the few tweaks to this PR before merging.

Issue #787, opened 2026-08-17
Enhancement: `SQLAlchemyStore` (recommended session store) has no way to purge expired rows


hasansezertasan:
### Summary

Docs recommend store-based sessions (`SQLAlchemyStore` under `stores={"sessions": ...}`). But `SQLAlchemyStore` cannot delete expired rows. Expiry lazy only: `get()`/`exists()` filter on `expires_at`, expired session reads absent, row never removed. Recommended path grows table unbounded. Only bulk op is `delete_all()`, kills live sessions too.

`delete_expired()` lives only on backends (`SQLAlchemyAsyncSessionBackend` / `SQLAlchemySyncSessionBackend`, `session.py`) — the backend-based path docs de-emphasize. Store users never build a backend, so docs cleanup line (`await session_backend.delete_expired()`) unreachable for them.

Raised before, unaddressed: Discussion #780, comment on PR #749. Not same as #745 (sessions not persisting, fixed by #749); here persist works, cleanup missing.

Ask: give store path a supported expired-row purge, matching backend.

### Basic Example

```python
import asyncio, datetime, aiosqlite
from advanced_alchemy.extensions.litestar import SQLAlchemyAsyncConfig
from advanced_alchemy.extensions.litestar.store import SQLAlchemyStore, StoreModelMixin

class S(StoreModelMixin):
    __tablename__ = "s"

cfg = SQLAlchemyAsyncConfig(connection_string="sqlite+aiosqlite:///exp.db", create_all=True)

async def main():
    async with cfg.get_engine().begin() as c:
        await c.run_sync(StoreModelMixin.metadata.create_all)
    store = SQLAlchemyStore(cfg, model=S, namespace="a")
    await store.set("k1", b"v", expires_in=datetime.timedelta(seconds=-10))  # already expired
    print("get:", await store.get("k1"))                            # -> None (lazily filtered)
    print("has delete_expired:", hasattr(store, "delete_expired"))  # -> False
    async with aiosqlite.connect("exp.db") as db:
        cur = await db.execute("SELECT count(*) FROM s")
        print("rows remaining:", (await cur.fetchone())[0])         # -> 1 (never reaped)

asyncio.run(main())
```

Output:

```
get: None
has delete_expired: False
rows remaining: 1
```

Wanted API mirrors backend: `await store.delete_expired()` running `DELETE ... WHERE expires_at <= now()`.

### Drawbacks and Impact

[...]

PR #788, opened 2026-08-17
docs(litestar): note SQLAlchemyStore has no expired-session cleanup

hasansezertasan:
## Description

Store-based session integration is the recommended path, but `SQLAlchemyStore` has no `delete_expired()`. Expiry lazy: `get()` filters expired entries, rows never removed, table grows unbounded. Docs cleanup guidance points at `session_backend.delete_expired()` — object store users never create.

This PR adds a note under the store-based section documenting the limitation plus a scheduled-cleanup workaround (`DELETE ... WHERE expires_at <= now()`). Docs-only; no code change.

Proper fix (adding `delete_expired()` to the store) tracked separately.

Refs #787

<hr>
📚 Documentation preview: <a href="https://litestar-org.github.io/advanced-alchemy-docs-preview/788">https://litestar-org.github.io/advanced-alchemy-docs-preview/788</a>

### DIFF
--- a/docs/usage/frameworks/litestar.rst
+++ b/docs/usage/frameworks/litestar.rst
@@ -624,8 +624,10 @@ The registered store name must match ``ServerSideSessionConfig.store``, which is
                 )
                 await db_session.commit()
 
-    Run it from an ``on_startup`` hook, a scheduler such as SAQ, or an external cron
-    job. Do not use ``delete_all()`` for this — it removes live sessions too.
+    Run it on a **recurring** schedule — a periodic task scheduler such as SAQ, or an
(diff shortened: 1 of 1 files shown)


docs/usage/frameworks/litestar.rst
```diff
@@ -603,6 +603,32 @@ Register a :class:`SQLAlchemyStore` under the session store name and let Litesta
 
 The registered store name must match ``ServerSideSessionConfig.store``, which is ``"sessions"`` unless you override it.
 
+.. note::
```
cofin, line 606: We should just add the job and document how to call it.  @hasansezertasan do you want to modify this to handle those changes?

PR #796, opened 2026-09-06
fix: pyright errors for ModelProtocol

### LINKED ISSUE #768: Bug: `ModelProtocol` rejects plain `DeclarativeBase` models under mypy — protocol members declared as instance variables, SQLAlchemy declares them `ClassVar`
### Description

`ModelProtocol`'s docstring says *"Any class with a mapper and table (including SQLModel `table=True` models) satisfies this protocol."* Under mypy this has never been true for models that don't inherit Advanced Alchemy's own bases. The protocol declares its members as instance variables:

```python
@runtime_checkable
class ModelProtocol(Protocol):
    if TYPE_CHECKING:
        __table__: FromClause
        __mapper__: Mapper[Any]
        __name__: str
```

while `sqlalchemy.orm.DeclarativeBase` declares all three as `ClassVar` (`sqlalchemy/orm/decl_api.py`). Mypy enforces that variance strictly, so every model built on plain `DeclarativeBase` fails the protocol — and with it, the `ModelT` bound on every repository and service.

### MCVE

```python
from advanced_alchemy.base import ModelProtocol
from advanced_alchemy.repository import SQLAlchemySyncRepository
from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column

class Base(DeclarativeBase): ...

class User(Base):
    __tablename__ = "user"
    id: Mapped[int] = mapped_column(primary_key=True)

x: ModelProtocol = User(id=1)

class UserRepository(SQLAlchemySyncRepository[User]):
    model_type = User
```

```
error: Incompatible types in assignment (expression has type "User", variable has type "ModelProtocol")  [assignment]
note: Protocol member ModelProtocol.__mapper__ expected instance variable, got class variable
note: Protocol member ModelProtocol.__name__ expected instance variable, got class variable
error: Type argument "User" of "SQLAlchemySyncRepository" must be a subtype of "ModelProtocol"  [type-var]
```

Runtime is fine (`isinstance(User(id=1), ModelProtocol)` is `True`); this is purely a static-typing defect.

### Why Advanced Alchemy's own bases pass

Only by MRO accident: `UUIDBase` lists `CommonTableAttributes` (whose `BasicAttributes` declares the members as instance variables) *before* `AdvancedDeclarativeBase`, so the instance-variable view wins the MRO. Writing `class Base(DeclarativeBase, BasicAttributes)` — `DeclarativeBase` first — reproduces the failure even with AA's own mixin. The `SQLModelBaseLike` placeholder in `_typing.py` has the same instance-variable declarations, so the SQLModel-compat claim also only holds through the placeholder, not for real models.

### Impact

Any project pointing AA repositories/services at an existing SQLAlchemy 2.0 model layer gets two `[type-var]` errors per repository/service pair. On our codebase that's 110 errors across ~55 service classes, which currently forces us to stub the entire package to `Any` and lose all of its types.

### Proposed fix

Declare the members as `ClassVar` in `ModelProtocol`, `BasicAttributes`, and `SQLModelBaseLike`, matching SQLAlchemy's own declarations. All three sites are `TYPE_CHECKING`-only, so runtime is unaffected; mypy and pyright accept both AA's own bases and plain declarative models afterwards. PR incoming.

Related: #234 asked to remove the `ModelT` bound entirely (closed) — this is a narrower fix that keeps the bound and makes the protocol match its documented contract.

**Versions:** advanced-alchemy 1.11.0 (reproduces on `main`), SQLAlchemy 2.0.x, mypy 1.20.2 / 2.2.0, pyright 1.1.409, Python 3.11.

s3rius:
## Description

This PR is intended to fix #768 regression. 

Since `__name__` is a special attribute defined for all `Type[T]` there is no need to declare it explicitly. And declaring it as a ClassVar breaks pyright and all LSP suggestions. 

I could come up with an MRE for this issue on pyright. 

```python
from typing import TYPE_CHECKING, ClassVar, Generic, Protocol, TypeVar


class DeclarativeMeta(type):
    # stands in for sqlalchemy.orm.decl_api.DeclarativeAttributeIntercept
    pass


class Base(metaclass=DeclarativeMeta):
    # stands in for sqlalchemy.orm.DeclarativeBase, which declares these
    if TYPE_CHECKING:
        __name__: ClassVar[str]


class Author(Base): ...


class ModelProtocol(Protocol):
    # stands in for advanced_alchemy.base.ModelProtocol
    if TYPE_CHECKING:
        __name__: ClassVar[str]


ModelT = TypeVar("ModelT", bound=ModelProtocol)


class Repository(Generic[ModelT]):
    # stands in for advanced_alchemy.repository.SQLAlchemySyncRepository
    ...


class AuthorRepo(Repository[Author]): ...
```

This script gives errors for pyright and passes with mypy.

```bash
# mypy --strict mre.py
Success: no issues found in 1 source file

# pyright mre.py 
/home/s3rius/projects/github/advanced-alchemy/mre.py
  /home/s3rius/projects/github/advanced-alchemy/mre.py:32:29 - error: Type "Author" cannot be assigned to type variable "ModelT@Repository"
    Type "Author" is not assignable to upper bound "ModelProtocol" for type variable "ModelT@Repository"
      "Author" is incompatible with protocol "ModelProtocol"
        "__name__" is defined as a ClassVar in protocol (reportInvalidTypeArguments)
1 error, 0 warnings, 0 informations
```

<hr>
📚 Documentation preview: <a href="https://litestar-org.github.io/advanced-alchemy-docs-preview/796">https://litestar-org.github.io/advanced-alchemy-docs-preview/796</a>

### DIFF
--- a/advanced_alchemy/base.py
+++ b/advanced_alchemy/base.py
@@ -156,7 +156,6 @@ class ModelProtocol(Protocol):
     if TYPE_CHECKING:
         __table__: ClassVar[FromClause]
         __mapper__: ClassVar[Mapper[Any]]
-        __name__: ClassVar[str]
 
 
 def model_to_dict(instance: "ModelProtocol", exclude: Optional[set[str]] = None) -> dict[str, Any]:


cofin (approved):
LGTM

